PR #24, opened 2020-10-07
Govern, turn into monorepo, prototype Discord bot

### LINKED ISSUE #28: Deprecate sync-abis script after merging erc3k
After #27 I believe we can just get rid of the script

### LINKED ISSUE #30: Fix CD
Works locally but on CD: https://rinkeby.etherscan.io/tx/0xe7f625ba36f45c84ad47243294bf0660e8b52cf5a0b60e52616334d55fb38b9d

Evalir:
Closes #30 , Closes #28


izqui:
Awesome, thanks Enrique. Couple comments:

- I think that eventually it might make sense to lift erc3kjs to the erc3k repo
- Aragon Govern is still not a final name, but let's leave for now. Also wonder if even if the actual package names are different, whether the dir structure in this repo should be `packages/contracts` or `packages/subgraph` cc @sohkai

Evalir:
> I think that eventually it might make sense to lift erc3kjs to the erc3k repo

From an offline conversation: we've settled into merging the ERC3k repo here and renaming this one, instead of the opposite.

> Aragon Govern is still not a final name, but let's leave for now. Also wonder if even if the actual package names are different, whether the dir structure in this repo should be packages/contracts or packages/subgraph

We could totally do it this way; I'm being a bit verbose with the naming right now. We could replicate the `aragon-apps` repo were it's just `apps/<app-name>`.

izqui (approved):
LGTM. Let's merge already if CI passes

izqui:
@Evalir CI should be good except for erc3k.js failing test. Pls fix and merge

Evalir:
Fixed; currently having a look at the eaglet deployment error.

izqui:
@evalir wanna try to do #28 before merging?

Evalir:
~How does merging ERC3K deprecate the script?~

EDIT: seems that The Graph plays nice with external ABIs from other packages 😄

izqui (approved):

Evalir:
🦅Let ten thousand DAOs bloom 🦅

PR #35, opened 2020-10-08
Review + refactor subgraph

0xGabi:
Leave a few comments to help with the review


packages/govern-subgraph/schema.graphql
```diff
@@ -1,22 +1,55 @@
-type Govern @entity {
+type ERC3000Registry @entity {
```
kanaagawa, line 1: I included the `ERC3000Registry` to keep track of every pair (dao, queue) registered under `name`.

packages/govern-subgraph/schema.graphql
```diff
@@ -1,22 +1,55 @@
...
+  id: ID!
   name: String!
-  queue: OptimisticQueue! @derivedFrom(field: "govern")
-  roles: [Role]!
+  dao: Govern!
+  queue: OptimisticQueue!
+  metadata: Bytes
+}
+
+type Govern @entity {
```
kanaagawa, line 16: I change quite a bit the `Govern` entity to focus on the executions it does as it's the main responsability of the Govern smartcontract. I also remove the link with the `ERC3000` (OptimisticQueue) entities to simplify and focus on the main functionality of the executions.

At one point I thought of including a `queues` attribute to keep track of every queue associated with the particular Govern. But I'm nos super sure how useful that might be.

packages/govern-subgraph/schema.graphql
```diff
@@ -1,22 +1,55 @@
...
+type Action @entity {
+  id: ID!
+  to: Bytes!
+  value: BigInt!
+  data: Bytes!
+  item: Item!
+  execution: Execution
 }
 
 type OptimisticQueue @entity {
```
kanaagawa, line 40: The `OptimisticQueue` change as well. I tried to decouple of the container concept and instead provide a more friendly and reach structure. Now we have the `items` on the queue (that are schedule) and then we also have `Challenge` items and `Veto` items.

I'm not convinced yet if we should also have the `Execution` here and how maybe a better way to link them with the `Govern` executions. This is one of the section that I'm unsure about the smartcontrats.

packages/govern-subgraph/src/ERC3000Registry.ts
```diff
@@ -0,0 +1,54 @@
...
+  GovernTemplate.create(event.params.dao);
+  QueueTemplate.create(event.params.queue);
+}
+
+export function handleSetMetadata(event: SetMetadataEvent): void {
+  // TODO:
+}
+
+
+export function loadOrCreateRegistry(
```
kanaagawa, line 43: I use the pattern `loadOrCreate` that Facu suggested on Connect and is great to simplify code and prevent inconsistencies with how The Graph process events.

packages/govern-subgraph/src/Govern.ts
```diff
@@ -0,0 +1,112 @@
...
+  // add the execution
+  const currentExecutions = govern.executions;
+  currentExecutions.push(execution.id);
+  govern.executions = currentExecutions;
+
+  execution.save();
+  govern.save();
+}
+
+// MiniACL Events
```
kanaagawa, line 31: Using the above pattern the logic for the ACL events was simplified quite a bit. I still want to double check if some of the syntax that I use (like mappings, filters, etc) has not problem during the subgraph deployment.

packages/govern-subgraph/src/OptimisticQueue.ts
```diff
@@ -9,210 +9,311 @@ import {
...
+  const currentItem = queue.queue;
+  currentItem.push(item.id);
+  queue.queue = currentItem;
+
+  item.save();
+  queue.save();
+}
+
+export function handleExecuted(event: ExecutedEvent): void {
+  // TODO: (Gabi) decide if we want to handle executinons here
```
kanaagawa, line 65: As I mentioned on a previous comments I was having a bit of dificulty linking the `Execution` between the Govern and the ERC3000 queue. I think we may need the `containerHash` on the `IERC3000` Executed method.

packages/govern-subgraph/src/OptimisticQueue.ts
```diff
@@ -9,210 +9,311 @@ import {
...
 
 export function handleChallenged(event: ChallengedEvent): void {
-  const container = Container.load(event.params.containerHash.toHexString())
-  container.executionState = CHALLENGED_TYPE
-  container.save()
+  const item = loadOrCreateItem(event.params.containerHash, event);
+
+  item.status = CHALLENGED_STATUS;
+
+  const challenge = loadOrCreateChallenge(item.id, event);
```
kanaagawa, line 79: I include the basic information of a challenge but we probably want to sync with @facuspagnuolo to make sure we are providing everything we need to interact with AP.

packages/govern-subgraph/src/lib/MiniACL.ts
```diff
@@ -0,0 +1,71 @@
...
+): RoleEntity {
+  const role = loadOrCreateRole(entity, selector, who);
+  role.granted = false;
+
+  role.save();
+
+  return role;
+}
+
+function buildRoleId(entity: Address, selector: Bytes, who: Address): string {
```
kanaagawa, line 40: Having this kind of helpers to help is build the id of an entity I found it super nice. Also a pro tip from Facu 😎

packages/govern-subgraph/src/lib/MiniACL.ts
```diff
@@ -0,0 +1,71 @@
+import { Address, Bytes } from "@graphprotocol/graph-ts";
+import { Role as RoleEntity } from "../../generated/schema";
+
+export function frozenRoles(roles: string[], selector: Bytes): void {
+  roles
```
kanaagawa, line 5: I decided to handle the frozen of roles in this way I think is similar to what you did, let me know what you think.

packages/erc3kjs/src/lib/abi/ERC3000Registry.json
```diff
@@ -1 +1,88 @@
...
\ No newline at end of file
+[
```
Evalir, line 1: Let's delete this file, and just reference the artifacts in the `govern-core` repo directly.

packages/erc3kjs/src/lib/abi/EagletFactory.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+[{"inputs":[{"internalType":"contract ERC3000Registry","name":"_registry","type":"address"},{"internalType":"contract OptimisticQueueFactory","name":"_queueFactory","type":"address"}],"stateMutability":"nonpayable","type":"constructor"},{"inputs":[{"internalType":"string","name":"_name","type":"string"}],"name":"newDummyGovern","outputs":[{"internalType":"contract Govern","name":"govern","type":"address"},{"internalType":"contract OptimisticQueue","name":"queue","type":"address"}],"stateMutability":"nonpayable","type":"function"},{"inputs":[],"name":"queueFactory","outputs":[{"internalType":"contract OptimisticQueueFactory","name":"","type":"address"}],"stateMutability":"view","type":"function"},{"inputs":[],"name":"registry","outputs":[{"internalType":"contract ERC3000Registry","name":"","type":"address"}],"stateMutability":"view","type":"function"}]
```
Evalir, line 1: Same as above

packages/erc3kjs/src/lib/abi/OptimisticQueueFactory.json
```diff
@@ -1 +1,58 @@
...
\ No newline at end of file
+[
```
Evalir, line 1: Ditto

packages/govern-subgraph/schema.graphql
Evalir: Note that names are unique, and you cannot have 2 DAOs with the same name. What you can have though, is DAOs registered multiple times, with multiple names. I'm expecting this to change per Facu's review though. 😄

packages/govern-subgraph/schema.graphql
Evalir: Yep, gotcha. I'd definitely keep the link with the `OptimisticQueue` as right now the thought process in the contracts is to have a Queue associated with a Govern entity, but of course this might change. We can keep it this way though if we're gonna use the `ERC3000Registry` to lookup DAOs, which _should_ always retrieve both the Govern entity and the Queue; if not, we could effectively have "headless" DAOs where we can only find its Govern entity, but not its Queue.

On the other side, I do like the change 😄 but I'm hesitant about the `dao` field as I don't know if we want to be so specific about the naming. Would love to hear @izqui thoughts on this

packages/govern-subgraph/schema.graphql
Evalir: I _definitely_ think executions should be here. The Queue's the main point of interaction with the DAO—we schedule, challenge and execute actions from it. If anything, we might just want to rethink how to link a Govern entity with its Queue entity to not have duplicate executions. 

Apart from this, love this change!

packages/govern-subgraph/src/ERC3000Registry.ts
Evalir: 💯 💥

packages/govern-subgraph/src/lib/MiniACL.ts
Evalir: Way better than what I did!

packages/govern-subgraph/src/lib/MiniACL.ts
Evalir: Yeah it's soooo clean! Facu tips are 🔥

packages/govern-subgraph/subgraph.template.yaml
```diff
@@ -2,7 +2,7 @@
 # Instead, edit subgraph.template.yaml and run `yarn manifest-<network>-[staging]`
 specVersion: 0.0.2
 description: A template for Aragon subgraphs
-repository: https://github.com/aragon/connect
+repository: https://github.com/aragon/govern
```
Evalir, line 5: Good catch 🙏

Evalir (approved):
This is looking 😍 just had some points to discuss. Could we deploy this subgraph on a separate instance to see how it would look? I might test both versions of it today with the Govern Console.

packages/govern-subgraph/schema.graphql
kanaagawa: Right, I used the`name` as the ID for each Entry entity (Govern, Queue) pair.

packages/govern-subgraph/schema.graphql
kanaagawa: Cool, yes I think that's the way to go, have `Govern` and `OptimisticQueue` linked at the `ERC3000Registry` entity and maybe a better name for the pair would be games as we can consider that each queue would be like the dao is playing a particular optimistic game?

On the naming I agree. Maybe also call it `executor`?

PR #42, opened 2020-10-09
Core, erc3k: inline comments

izqui:


packages/erc3k/contracts/IERC3000.sol
```diff
@@ -9,21 +9,53 @@ import "./ERC3000Data.sol";
...
+       the current configuration of the system
+     * @param resolverId disputeId in the arbitrator in which the dispute over the container was created
+     */
     function resolve(ERC3000Data.Container memory container, uint256 resolverId) virtual public returns (bytes[] memory execResults);
     event Resolved(bytes32 indexed containerHash, address indexed actor, bool approved);
 
+    /**
+     * @notice Apply arbitrator's ruling over a challenge once it has come to a final ruling
+     * @param payloadHash Hash of the payload being vetoed
+     * @param config A ERC3000Data.Config struct holding the config attached to the payload being vetod
```
kanaagawa:
```suggestion
     * @param config A ERC3000Data.Config struct holding the config attached to the payload being vetoed

File: packages/erc3k/contracts/IERC3000.sol [hunk 2]
```diff
@@ -9,21 +9,53 @@ import "./ERC3000Data.sol";
...
+     */
     function veto(bytes32 payloadHash, ERC3000Data.Config memory config, bytes memory reason) virtual public;
     event Vetoed(bytes32 indexed containerHash, address indexed actor, bytes reason, ERC3000Data.Collateral collateral);
 
+    /**
+     * @notice Apply a new configuration for all *new* containers to be scheduled
+     * @param config A ERC3000Data.Config struct holding all the new params that will control the queue
```
kanaagawa: Maybe instead of `queue` keep using the `system` terminology as above?

packages/govern-core/contracts/OptimisticQueue.sol
```diff
@@ -14,115 +14,157 @@ import "./lib/OptimisticQueueStateLib.sol";
...
     }
 
-    function resolve(ERC3000Data.Container memory _container, uint256 _resolverId) override public returns (bytes[] memory execResults) {
+    /**
+     * @notice Apply arbitrator's ruling over a challenge once it has come to a final ruling
+     * @param _container A ERC3000Data.Container struct holding both the paylaod being scheduled for execution and
+       the current configuration of the system
+     * @param _disputeId disputeId in the arbitrator in which the dispute over the container was created
+     */
+    function resolve(ERC3000Data.Container memory _container, uint256 _disputeId) override public returns (bytes[] memory execResults) {
```
kanaagawa, line 151: I notice we changed `_disputedId` here but the `IERC3000` is still `resolverId`. Should we change that as well?

packages/govern-core/contracts/OptimisticQueue.sol
```diff
@@ -14,115 +14,157 @@ import "./lib/OptimisticQueueStateLib.sol";
...
 
         bytes32 containerHash = _container.hash();
         queue[containerHash].checkAndSetState(
-            OptimisticQueueStateLib.State.Scheduled,
+            OptimisticQueueStateLib.State.Scheduled, // note that we will revert here if the container wasn't previously scheduled
             OptimisticQueueStateLib.State.Executed
         );
 
-        _container.config.scheduleDeposit.releaseTo(_container.payload.submitter);
+        _container.config.scheduleDeposit.releaseTo(_container.payload.submitter); // release collateral to executor
```
kanaagawa, line 105: I wonder why we can call `releaseTo` method here without doing similar to what we are doing for `collectFrom`:
```
 ERC3000Data.Collateral memory collateral = _container.config.scheduleDeposit;
 collateral.collectFrom(_container.payload.submitter);
```

0xGabi (commented):
Much better with comments!

packages/erc3k/contracts/IERC3000.sol
izqui:
```suggestion
     * @param config A ERC3000Data.Config struct holding all the new params that will control the system
```

packages/govern-core/contracts/OptimisticQueue.sol
izqui: I don't understand the question

packages/govern-core/contracts/OptimisticQueue.sol
izqui: I am not sure tbh. This is a weird balance

PR #52, opened 2020-10-14
📦 NEW: generate abi files using buidler-extract

0xGabi:
Update how we generate the abis reusing the package we had [buidler-extract](https://github.com/0xGabi/buidler-extract). Fork of [truffle-extract](https://github.com/sohkai/truffle-extract) modified for buidler.


packages/govern-core/contracts/ERC3000Registry.sol
```diff
@@ -12,28 +12,28 @@ contract ERC3000Registry {
 
     mapping (string => bool) public nameUsed;
 
-    event Registered(ERC3000Executor indexed dao, ERC3000 queue, address indexed registrant, string name);
-    event SetMetadata(ERC3000Executor indexed dao, bytes metadata);
+    event Registered(ERC3000Executor indexed executor, ERC3000 queue, address indexed registrant, string name);
```
Evalir, line 15: Oooh interesting, are we changing these based on Facu's review?

Evalir (approved):
💯 💯

nivida (approved):

packages/govern-core/contracts/ERC3000Registry.sol
facuspagnuolo: My point on the registry was that it is linking two things (queue and executor) that may not be used in that way and indeed it can be misleading cause anyone can register that pair allowing to be overwritten overtime. We need to think a bit more about how this registry is supposed to be used, or maybe I'm missing something

facuspagnuolo: I misunderstood your comment @Evalir, sorry about that, the rename looks fine. However, my comment about the registry I think it still stands

kanaagawa: Yes I did it because of Facu's comment. I think the Registry is important because is the way we link a Queue (Optimistic Game) with an Execturor (DAO). I agree with Facu's point that is still unclear the role of the Registry other than as a way of link the pair off chain.

Evalir: Definitely agree about the Registry; I've been treating this particular contract as very WIP and very prone to change. Mainly because of this:

A Govern DAO is composed of two things (potentially more, but this is the core):
- Executor (Govern)
- OptimisticQueue.

I interpret it as if there _should be_ a hard dependency between these. You could say that the equivalent of installing another "executor" to a DAO is just changing the executor in the container, but this conflicts with the above. As the executor also holds funds, I feel we should not be so lightweight in terms of being able to change executors, or people might just lose track of funds at some point if the pair gets overwritten. 

I don't exactly know what the best way going forward might be, but I think it makes the case for either enforcing a dependency at the code level, or a separate vault contract. I'd go with the first one, as I see an executor as an Agent.

PR #55, opened 2020-10-15
Adds the development environment for govern-server

nivida:
Created a docker-compose and updated the ``.gitignore`` file.


scripts/init.dev.env.sh
```diff
@@ -0,0 +1,118 @@
...
+    echo -e ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>${NC}"
+
+    exit 1
+fi
+
+# --------------------------------------------
+# Wait some seconds until containers are ready
+# --------------------------------------------
+echo "Waiting until containers are ready..."
+sleep 35
```
nivida: This timeout can later get removed but it does its job for the time being (MVP state).

packages/govern-server/package.json
```diff
@@ -4,7 +4,8 @@
   "license": "GPL-3.0",
   "main": "src/index.ts",
   "scripts": {
-    "start": "node --loader ts-node/esm.mjs  src/index.ts"
+    "start": "yarn start:containers && node --loader ts-node/esm.mjs  src/index.ts",
```
bpierre, line 7: I think this should work now:

```suggestion
    "start": "yarn start:containers && ts-node src/index.ts",
```

(I started using `node --loader ts-node/esm.mjs` when I was having troubles to load the modules, but I noticed later that it was working fine with `ts-node` directly.)

scripts/init.dev.env.sh
nivida: Improved.

bpierre (approved):
:fire: :fire: :fire:

packages/govern-server/package.json
nivida: Yup.. kept the command as it is cause I saw the change in your PR :-)

nivida:
@bpierre Thanks for quickly doing those discussed changes! :-)

PR #57, opened 2020-10-15
👌 IMPROVE: handle Govern metadata + include ERC20 entity

0xGabi:


packages/govern-subgraph/src/OptimisticQueue.ts
```diff
@@ -11,309 +11,320 @@ import {
...
+const VETOED_STATUS = 'Vetoed'
+
+const ALLOW_RULING = BigInt.fromI32(4)
 
 export function handleScheduled(event: ScheduledEvent): void {
-  const queue = loadOrCreateQueue(event.address);
+  const queue = loadOrCreateQueue(event.address)
 
-  const item = loadOrCreateItem(event.params.containerHash, event);
+  const item = loadOrCreateItem(event.params.containerHash, event)
```
Evalir, line 42: I'm under the slight impression that is isn't working properly. AFAIK the latest subgraph version running is from this PR, and I've sent a few actions to a test DAO (`bitconnectr`) and it isn't picking up the events.

packages/govern-subgraph/src/OptimisticQueue.ts
```diff
@@ -11,309 +11,320 @@ import {
...
-  let item = ItemEntity.load(itemId);
+  let item = ItemEntity.load(itemId)
   if (item === null) {
-    item = new ItemEntity(itemId);
-    item.status = NONE_STATUS;
-    item.actions = [];
-    item.createdAt = event.block.timestamp;
+    item = new ItemEntity(itemId)
+    item.status = NONE_STATUS
+    item.actions = []
```
Evalir: I think we're not handling this `actions` array anywhere else.

Evalir (commented):
Looking good! A few comments, and I think we should double check the logic for indexing actions (items, or `Containers` in the ERC3000 standard)

packages/govern-subgraph/src/OptimisticQueue.ts
kanaagawa: Actually we should remove this line because the relation is derived from each `Action` entity.

packages/govern-subgraph/src/OptimisticQueue.ts
kanaagawa: You are right, I'll go ahead and merge this changes and then review the realation in adition of the new changes of the contracts given that a few links that were missing and maybe related

PR #58, opened 2020-10-15
👌 IMPROVE: move govern-core to typescript

0xGabi:
Move to use buidler with typescript. 
Use `buidler-local-networks-config-plugin` to handle network configuration with a local file.

TODO:
- [ ] Smart contract type generation using typechain. Several of the contract are generated but we are getting an error on `DepositLib`. I wonder if may be related with using the ABIEncoderV2


packages/govern-core/abi/ERC3000Registry.json
```diff
@@ -73,7 +73,7 @@
       "inputs": [
         {
           "internalType": "contract ERC3000Executor",
-          "name": "_dao",
+          "name": "_executor",
```
Evalir, line 76: I notice some of these are using `_executor` and others are using `executor`; are these correct?

packages/govern-core/buidler.config.ts
```diff
@@ -0,0 +1,156 @@
...
+    )
+    print(governFactory.address, 'GovernFactory')
+
+    if (process.env.CD) {
+      writeFileSync(FACTORY_CACHE_NAME, governFactory.address)
+    }
+  }
+)
+
+task('deploy-govern', 'Deploys an Govern from provided factory')
```
Evalir, line 75: I've had this on my mind for a while... Should we move these to separate files? I feel this file is crowded.

packages/govern-core/buidler.config.ts
```diff
@@ -0,0 +1,156 @@
...
+    print(queue, 'Queue')
+  })
+
+const config: BuidlerConfig = {
+  // This is a sample solc configuration that specifies which version of solc to use
+  solc: {
+    version: '0.6.8',
+    optimizer: {
+      enabled: true,
+      runs: 2000 // TODO: target average DAO use
```
Evalir, line 134: Should we do this now?

packages/govern-core/buidler.config.ts
```diff
@@ -0,0 +1,156 @@
...
+} from 'unique-names-generator'
+
+usePlugin('buidler-local-networks-config-plugin')
+usePlugin('buidler-typechain')
+usePlugin('solidity-coverage')
+usePlugin('@nomiclabs/buidler-ethers')
+usePlugin('@nomiclabs/buidler-etherscan')
+usePlugin('@nomiclabs/buidler-waffle')
+
+dotenv.config({ path: '../../.env' })
```
Evalir, line 19: Just noticed: this is getting the `.env` file from the root folder, while the above `.env.example` is on this folder. Should we move the `.env.example` to the root folder?

Evalir (commented):
Looking good 😄 some comments:

Evalir:
Also, just noticed this comment: 

> Smart contract type generation using typechain. Several of the contract are generated but we are getting an error on DepositLib. I wonder if may be related with using the ABIEncoderV2

It might be the case that it's the `ABIEncoderV2`; I've been battling against a frontend issue + testing due to the encoder misbehaving. I will file an issue on Ethers soon.

Evalir:
Ah, @0xGabi seems that #66 makes this PR obsolete; it makes more sense to work on top of this new PR, or wait for it to be merged. Should we close this and work on top of the new one?

0xGabi:
Yes probably most of changes needs to rebase to #66. I'll close this one and keep the branch for reference

PR #59, opened 2020-10-16
Server: add wrapper to fetch data from the graph

0xGabi:
Reuse the scheme we were using on Connect to have a basic wrapper that only supports query data from the graph.

### DIFF
--- a/packages/govern-server/src/core/data/TheGraphWrapper.ts
+++ b/packages/govern-server/src/core/data/TheGraphWrapper.ts
@@ -1,0 +1,70 @@
+import fetch from 'isomorphic-unfetch'
+import { Client } from '@urql/core'
+import { DocumentNode } from 'graphql'
+import { ErrorConnection } from '../errors'
+import { QueryResult } from '../types'
+
+type TheGraphWrapperOptions = {
+  verbose?: boolean
+}
+
+export default class TheGraphWrapper {
+  #client: Client
+  #verbose: boolean
+
+  constructor(
+    subgraphUrl: string,
+    options: TheGraphWrapperOptions | boolean = {}
+  ) {
+    if (typeof options === 'boolean') {
+      console.warn(
+        'TheGraphWrapper: please use `new TheGraphWrapper(url, { verbose })` rather than `new TheGraphWrapper(url, verbose)`.'
+      )
+      options = { verbose: options }
+    }
+    options = options as TheGraphWrapperOptions
+
+    this.#verbose = options.verbose ?? false
+    this.#client = new Client({ maskTypename: true, url: subgraphUrl, fetch })
+  }
+
+  async performQuery(
+    query: DocumentNode,
+    args: any = {}
+  ): Promise<QueryResult> {
+    const result = await this.#client.query(query, args).toPromise()
+
+    if (this.#verbose) {
+      console.log(this.describeQueryResult(result))
+    }
+
+    if (result.error) {
+      throw new ErrorConnection(
+        this.describeQueryResultError(result) + this.describeQueryResult(result)
+      )
+    }
+
+    return result
+  }
+
+  private describeQueryResult(result: QueryResult): string {
+    const queryStr = result.operation.query.loc?.source.body
+    const dataStr = JSON.stringify(result.data, null, 2)
+    const argsStr = JSON.stringify(result.operation.variables, null, 2)
+    const subgraphUrl = result.operation.context.url
+
+    return (
+      `Subgraph: ${subgraphUrl}\n\n` +
+      `Arguments: ${argsStr}\n\n` +
+      `Query: ${queryStr}\n\n` +
+      `Returned data: ${dataStr}\n\n`
+    )
+  }
+
+  private describeQueryResultError(result: QueryResult): string {
+    if (!result.error) {
+      return ''
+    }
+    return `${result.error.name}: ${result.error.message}\n\n`
+  }
+}
--- a/packages/govern-server/src/core/data/index.ts
+++ b/packages/govern-server/src/core/data/index.ts
@@ -1,6 +1,14 @@
 import { ErrorInvalidNetwork, ErrorUnexpectedResult } from '../errors'
 import { toNetwork } from '../utils'
-import { Address, Network, Networkish, QueryResult } from '../types'
+import {
+  Address,
+  DaoData,
+  Network,
+  Networkish,
+  OptimisticGameData,
+  OptimisticQueueData,
+  QueryResult,
+} from '../types'
 import * as queries from './queries'
 import TheGraphWrapper from './TheGraphWrapper'
 
@@ -36,7 +44,8 @@ class ConnectorTheGraph {
 
     if (!orgSubgraphUrl) {
       throw new ErrorInvalidNetwork(
-        `The chainId ${this.network.chainId} is not supported by the TheGraph connector.`
+        `The chainId ${this.network.chainId} is not supported ` +
+          `by the TheGraph connector.`
       )
     }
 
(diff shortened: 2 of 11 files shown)


packages/govern-server/src/core/data/TheGraphWrapper.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  async performQuery(
+    query: DocumentNode,
+    args: any = {}
+  ): Promise<QueryResult> {
+    const result = await this.#client.query(query, args).toPromise()
+
+    if (this.#verbose) {
+      console.log(this.describeQueryResult(result))
```
nivida, line 38: ``this.describeQueryResult(result)`` could get called multiple times. I think it would be smart to store it in a ``const``.

btw.: the method name ``mapQueryResult`` is I think more self-explaining.

packages/govern-server/src/core/data/TheGraphWrapper.ts
```diff
@@ -0,0 +1,70 @@
...
+      )
+      options = { verbose: options }
+    }
+    options = options as TheGraphWrapperOptions
+
+    this.#verbose = options.verbose ?? false
+    this.#client = new Client({ maskTypename: true, url: subgraphUrl, fetch })
+  }
+
+  async performQuery(
```
nivida, line 31: Not sure how the standard is here. But is only ``private`` and ``protected`` used to define the scope of a class method or do we usually also add the ``public`` keyword? :-)

packages/govern-server/src/core/data/TheGraphWrapper.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (result.error) {
+      throw new ErrorConnection(
+        this.describeQueryResultError(result) + this.describeQueryResult(result)
+      )
+    }
+
+    return result
+  }
+
+  private describeQueryResult(result: QueryResult): string {
```
nivida, line 50: I would rename those two functions to ``map*`` and I think it isn't required to initiate those ``const``'s

packages/govern-server/src/core/data/TheGraphWrapper.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export default class TheGraphWrapper {
+  #client: Client
+  #verbose: boolean
+
+  constructor(
+    subgraphUrl: string,
+    options: TheGraphWrapperOptions | boolean = {}
+  ) {
+    if (typeof options === 'boolean') {
```
nivida, line 19: This ``options`` parsing logic could be moved out to a private ``parseOptions`` or ``mapOptions`` method. It would just be cleaner to not have any logic directly in the constructor of a class.

packages/govern-server/src/core/data/index.ts
```diff
@@ -0,0 +1,127 @@
...
+      })
+    } catch (err) {
+      throw new ErrorUnexpectedResult(
+        `Unexpected result when fetching the queues for dao ${dao}.`
+      )
+    }
+  }
+}
+
+export default ConnectorTheGraph
```
nivida, line 133: Instead of having it at the end of the file could we also: ``export default class ConnectorTheGraph`` 

... not sure what the standard is :-)

packages/govern-server/src/core/params.ts
```diff
@@ -0,0 +1,17 @@
+export const NETWORKS = [
```
nivida, line 1: This could probably be within a ``config.json`` file. I think it would also be more self-explaining instead of calling this file ``params``.

packages/govern-server/src/core/data/TheGraphWrapper.ts
```diff
@@ -0,0 +1,70 @@
...
+
+type TheGraphWrapperOptions = {
+  verbose?: boolean
+}
+
+export default class TheGraphWrapper {
+  #client: Client
+  #verbose: boolean
+
+  constructor(
```
nivida, line 15: Do we normally not add function documentations? Could be interesting later to generate a class reference :-)

packages/govern-server/src/core/data/index.ts
```diff
@@ -0,0 +1,127 @@
...
+  if (network.chainId === 4) {
+    return 'https://api.thegraph.com/subgraphs/name/0xgabi/aragon-govern-rinkeby-staging'
+  }
+  if (network.chainId === 100) {
+    return 'placeholder'
+  }
+  return null
+}
+
+class ConnectorTheGraph {
```
nivida, line 34: I think the naming convention changed here. Shouldn't it be called ``TheGraphConnector`` and I even wouldn't it call ``Connector`` we could name it ``TheGraphRepository`` or ``TheGraphServiceClient`` this cause the ``Wrapper`` is the actual "connector" 🤷‍♂️

packages/govern-server/src/core/data/TheGraphWrapper.ts
```diff
@@ -0,0 +1,70 @@
...
+import { Client } from '@urql/core'
+import { DocumentNode } from 'graphql'
+import { ErrorConnection } from '../errors'
+import { QueryResult } from '../types'
+
+type TheGraphWrapperOptions = {
+  verbose?: boolean
+}
+
+export default class TheGraphWrapper {
```
nivida, line 11: A name like ``TheGraphAdapter`` or ``TheGraphClient`` could be more self-describing. 🤷‍♂️

packages/govern-server/src/core/data/queries/fragments.ts
```diff
@@ -0,0 +1,133 @@
+import gql from 'graphql-tag'
```
nivida, line 1: Not sure if this dependency is required cause ``apollo-server`` does expose this dependency as well.. [see](https://github.com/aragon/govern/blob/master/packages/govern-server/src/api-graphql/index.ts#L1)..

packages/govern-server/src/core/errors.ts
```diff
@@ -0,0 +1,139 @@
+function defineNonEnumerable(instance: object, name: string, value: any) {
```
nivida, line 1: Could get added as ``private`` method to the ``ErrorException`` object cause it isn't used anywhere else and would force us if required to create a ``AbstractError`` class later. 🤷‍♂️

packages/govern-server/src/core/utils/network.ts
```diff
@@ -0,0 +1,84 @@
+import { ErrorInvalidNetwork } from '../errors'
```
nivida, line 1: Not sure how the structure is but we will use I think anyways all of those functions here. This means it will get bundled anyways and we could forward the creation of nice TS classes. 

What do you think about creating out of it a value object class?

``` typescript
const network = new Network('mainnet');
console.log(network.chainId);

const network = new Network(netObj);
console.log(network.chainId);

console.log(network.toObject());
> { ... }
```

packages/govern-server/src/core/data/index.ts
```diff
@@ -0,0 +1,127 @@
...
+import * as queries from './queries'
+import TheGraphWrapper from './TheGraphWrapper'
+
+export type ConnectorTheGraphConfig = {
+  network: Networkish
+  subgraphUrl?: string
+  verbose?: boolean
+}
+
+function getSubgraphUrl(network: Network): string | null {
```
nivida, line 21: This can be a private method of ``ConnectorTheGraph``

nivida (commented):
Added some comments. I think many of them are just popped up cause I'm not sure what the standards are here at Aragon. 

Overall: I would probably combine those two ``TheGraph`` classes to a ``TheGraphServiceClient``. 🤷‍♂️

packages/govern-server/src/core/data/TheGraphWrapper.ts
bpierre: We generally don’t use `public`, as it is the default − and it is more concise :smile:

packages/govern-server/src/core/data/index.ts
bpierre: I think this class could be the Govern JS API already? It would fetch data from The Graph (`TheGraphWrapper`), IPFS and other sources. We could move it one level up to be in `core/`, as it would be the main entry point to the core library.

packages/govern-server/src/core/data/index.ts
bpierre: We usually export at the end of files, it makes it easy to spot and avoids splitting the class declaration over two lines :smile:

packages/govern-server/src/core/data/queries/fragments.ts
bpierre: I think we should keep everything in `core/` separate from `apollo-server`, as this is the part that is going to move in a separate package and run in the browser eventually.

packages/govern-server/src/core/data/index.ts
nivida: This has to be a really really long class name lol.. But great to know! :-)

... This means the export has to be changed for the other TS class. :-)

packages/govern-server/src/core/data/queries/fragments.ts
nivida: I think then should we probably move the core out in a separate package if I got you right?

btw.: I was only seeing here the duplicated dependency ``graphql-tag``. 🤷‍♂️

packages/govern-server/src/core/data/index.ts
bpierre: > This has to be a really really long class name lol.. But great to know! :-)

haha yes, but I was more thinking of a full declaration with `implements` / `extends`, in which case it can exceed 80 chars.

packages/govern-server/src/core/data/queries/fragments.ts
bpierre: > I think then should we probably move the core out in a separate package if I got you right?

That’s exactly the idea: starting with `core/` in this package so we don’t spend any time on this separation at first, then move it to its own package to separate the server app from the core library.

nivida: As discussed on the call: I agree to move the core out of the server package as well because of the fully decentralized version of govern in the future. Means we can keep this dependency for now.

packages/govern-server/src/core/data/TheGraphWrapper.ts
nivida: Hint: As discussed with Pierre should the export of a class be on the last line of a file. :-)

Issue #49, opened 2020-10-12
Govern Server − Phase 4


bpierre:
Add the following features to the Govern API server:

- New frontend: REST API (might be a light, read-only version).
- Complete logs.
- Excellent self deploying story.

bpierre:
Closing for https://github.com/aragon/govern/issues/222

PR #60, opened 2020-10-16
GraphQL API

bpierre:
Available queries:

```gql
type Query {
  dao(address: String!): Dao!
  daos: [Dao!]!
  game(name: String!): OptimisticGame!
  games: [OptimisticGame!]!
}
```

Example to query a DAO and its queues:

```gql
dao {
  id
  queues {
    id
  }
}
```

Types:

```gql
type ERC3000Registry {
  id: ID!
  address: String!
  count: Int!
  games: [OptimisticGame!]
}

type OptimisticGame {
  id: ID!
  name: String!
  executor: Dao!
  queue: OptimisticQueue!
  metadata: String
}

type Dao {
  id: ID!
  address: String!
  games: [OptimisticGame!]
  executions: [Execution!]
  roles: [Role!]
  queues: [OptimisticQueue!]
}

type Execution {
  id: ID!
  sender: String!
  actions: [Action!]
  results: [String!]
}

type Action {
  id: ID!
  to: String!
  value: String!
  data: String!
  item: Item!
  execution: Execution
}

type OptimisticQueue {
  id: ID!
  address: String!
  config: Config
  games: [OptimisticGame!]
  queue: [Item!]
  executions: [Execution!]
  challenges: [Challenge!]
  vetos: [Veto!]
  roles: [Role!]
}

type Config {
  id: ID!
  queue: OptimisticQueue!
  executionDelay: String!
  scheduleDeposit: Collateral!
  challengeDeposit: Collateral!
  vetoDeposit: Collateral!
  resolver: String!
  rules: String!
}

type Collateral {
  id: ID!
  token: String!
  amount: String!
}

type Item {
  id: ID!
  status: ItemStatus!
  nonce: String!
  executionTime: String!
  submitter: String!
  executor: Dao!
  actions: [Action!]
  proof: String!
  collateral: Collateral!
  createdAt: String!
}

type Challenge {
  id: ID!
  queue: OptimisticQueue!
  challenger: String!
  item: Item!
  arbitrator: String!
  disputeId: String!
  evidences: [Evidence!]
  collateral: Collateral!
  ruling: String
  approved: Boolean
  createdAt: String!
}

type Evidence {
  id: ID!
  challenge: Challenge!
  data: String!
  submitter: String!
  createdAt: String!
}

type Veto {
  id: ID!
  queue: OptimisticQueue!
  item: Item!
  reason: String!
  submitter: String!
  collateral: Collateral!
  createdAt: String!
}

type Role {
  id: ID!
  entity: String!
  selector: String!
  who: String!
  granted: Boolean!
  frozen: Boolean!
}

enum ItemStatus {
  None
  Approved
  Cancelled
  Challenged
  Executed
  Rejected
  Scheduled
  Vetoed
}
```


0xGabi (approved):
The changes look good to me! I like the simple API you expose on the server.

packages/govern-server/package.json
```diff
@@ -17,7 +17,8 @@
     "ts-node": "^9.0.0"
   },
   "devDependencies": {
-    "typescript": "^3.8.3",
-    "@types/express": "^4.17.8"
+    "@types/express": "^4.17.8",
+    "ts-node-dev": "^1.0.0",
+    "typescript": "^4.0.3"
```
nivida: We shouldn't forget to keep the TS versions anywhere in-sync. Look for example at the root ``package.json``.

packages/govern-server/src/api-graphql/resolvers.ts
```diff
@@ -0,0 +1,28 @@
+import { IResolvers } from 'apollo-server'
+import GovernData from '../core/data'
+
+export default function resolvers(govern: GovernData): IResolvers {
+  return {
+    Query: {
+      async dao(_, args) {
+        return govern.dao(args.address)
```
nivida, line 8: Noice! The idea of the "API abstraction layer" looks really clean in his first draft/version 💪

packages/govern-server/src/core/data/index.ts
```diff
@@ -54,78 +53,90 @@ class ConnectorTheGraph {
     })
   }
 
-  private async fetchResult<T>(
+  private async fetchResult<ReturnType>(
```
nivida: Cause ``fetchResult`` is anyways a ``async`` method should we remove the ``callback`` which is a legacy pattern and just use the async/await construct :-)

packages/govern-server/src/env.ts
```diff
@@ -1,4 +1,14 @@
-export function graphqlHttpPort(): number {
+function graphqlHttpPort(): number {
```
nivida, line 1: Instead of using environment variables could we do it with configuration objects. Because with it are those helpers wouldn't be required.

packages/govern-server/src/index.ts
```diff
@@ -1,11 +1,17 @@
...
+  const govern = new GovernData({ network: 4 })
+
+  const { url: graphqlUrl } = await startGraphql({
+    govern,
+    httpPort: env.graphqlHttpPort,
+  })
+
   console.log()
-  console.log(`GraphQL API listening on ${url}`)
+  console.log(`GraphQL API listening on ${graphqlUrl}`)
```
nivida, line 14: Just a side-note: Is it planned to use an advanced logging mechanism in the final product as for example the Winston logger?

packages/govern-server/src/types.ts
```diff
@@ -0,0 +1 @@
+export type Address = string
```
nivida, line 1: Not sure if this type is required if we do not any further parsing of the passed URL 🤷‍♂️

packages/govern-server/src/core/data/index.ts
nivida: side-note: Is it a standard here at Aragon to not name the file as the class within is named?

bpierre: Yes I realize now that it’s not needed. It started as a “filtering / defaulting function” but it’s not doing anything special now.

packages/govern-server/package.json
bpierre: Yes good point, I’ll revert this and bump everything in a separate PR.

packages/govern-server/src/core/data/index.ts
bpierre: We usually do it except in certain cases, and I think we should do it here :smiley:

packages/govern-server/src/env.ts
bpierre: I like being able to configure the server without touching any file. Or do you see a way to have both?

packages/govern-server/src/index.ts
bpierre: Definitely yes. I put having “complete logs” in the [phase 4](https://github.com/aragon/govern/issues/49), but we could start having some logging mechanism before that.

packages/govern-server/src/types.ts
bpierre: It’s just for reading / documentation purposes, since we use those so much :slightly_smiling_face:

packages/govern-server/package.json
Evalir: Wonder if there's a "nicer" way to keep this all in sync?

bpierre: `yarn upgrade-interactive --latest` should be enough I think? It groups all the workspaces dependencies, so we can select all the `typescript` for example, and they will upgrade at once.

packages/govern-server/src/core/data/index.ts
bpierre: Done in https://github.com/aragon/govern/pull/60/commits/0ed5daa48b442e74106fb9234c0a7652fd9e19e3 and https://github.com/aragon/govern/pull/60/commits/41c5103c92194ffe92670270173cc8f87ffc5a04

packages/govern-server/src/env.ts
nivida: Not sure if actually was this tool I thought to use for web3.js but it was definitely something similar: https://github.com/commenthol/configg

... with such config management would you just have one environment variable you set.. for example ``production`` or ``local``

packages/govern-server/src/index.ts
nivida: Noice! :)

packages/govern-server/src/types.ts
nivida: Yup, true that.. this will definitely increase the readability.

packages/govern-server/src/env.ts
nivida: ... but we can change this later after ``Phase 4``

nivida (approved):

PR #64, opened 2020-10-19
Add @aragon/govern-contract-utils package

izqui:
Split contract utils into new package

**Intentionally breaks CI/CD which is fixed in #66**


packages/govern-contract-utils/scripts/sample-script.js
```diff
@@ -0,0 +1,29 @@
+// We require the Buidler Runtime Environment explicitly here. This is optional
```
Evalir, line 1: We could probably delete this file.

packages/govern-contract-utils/.gitattributes
```diff
@@ -0,0 +1 @@
+*.sol linguist-language=Solidity
```
kanaagawa, line 1: I think GitHub now supports Solidity by default.

0xGabi (approved):
LGTM! 👍

PR #68, opened 2020-10-19
core: optionality for allowing failures and add execution memo

izqui:
@0xGabi feel free to push the required subgraph changes on top of this branch

**CI failing due to an issue with the factory, to be fixed in a separate PR solving #36**


packages/govern-core/contracts/Govern.sol
```diff
@@ -23,18 +27,22 @@ contract Govern is ERC3000Executor, ACL {
...
+    function exec(ERC3000Data.Action[] memory actions, bytes32 allowFailuresMap, bytes32 memo) override public auth(EXEC_ROLE) returns (bytes[] memory) {
+        require(actions.length <= MAX_ACTIONS, "govern: too many"); // need to limit since we use 256-bit bitmaps
+        
         bytes[] memory execResults = new bytes[](actions.length);
+        bytes32 failureMap = BitmapLib.empty; // start with an empty bitmap
 
-        for (uint256 i = 0; i < actions.length; i++) {
-            // todo: optimize with assembly
+        for (uint8 i = 0; i < actions.length; i++) { // can use uint8 given the action limit
+            // TODO: optimize with assembly
```
Evalir, line 37: Should we do this anytime soon? 😄

packages/govern-contract-utils/contracts/bitmaps/BitmapLib.sol
```diff
@@ -0,0 +1,19 @@
+/*
+ * SPDX-License-Identifier:    GPL-3.0
+ */
+
+pragma solidity 0.6.8;
+
+library BitmapLib {
```
Evalir, line 7: Pretty nifty! Should we add some docs on how this works? I don't expect most frontend developers to be very familiar with bitwise operations.

packages/erc3k/contracts/IERC3000Executor.sol
```diff
@@ -8,7 +8,7 @@ pragma experimental ABIEncoderV2;
 import "./ERC3000Data.sol";
 
 abstract contract IERC3000Executor {
-    function exec(ERC3000Data.Action[] memory actions) virtual public returns (bytes[] memory);
+    function exec(ERC3000Data.Action[] memory actions, bytes32 allowFailuresMap, bytes32 memo) virtual public returns (bytes[] memory);
```
Evalir: Should we just call `allowFailuresMap` `failureMap`?

Evalir (commented):
Looking good! Had just a few comments.

Also, note this might be a breaking change for the subgraph due to the event changes and we should update it accordingly. (cc @0xGabi)

packages/govern-core/contracts/Govern.sol
izqui: Unsure if it is worth it, we should do some gas profiling when it is time for fine polishing, we are not there just yet

packages/govern-contract-utils/contracts/bitmaps/BitmapLib.sol
izqui: That'd be great. I think we should have an abstraction similar to how we do in the tests (an array of bools on whether things can fail which can be converted, more efficiently, to the bitmap)

packages/erc3k/contracts/IERC3000Executor.sol
izqui: No, could be confusing with the failure results bitmap which is logged (and actually should be returned here too)

izqui:
**Added some more breaking changes in de1ef89, 27b39ec, c3c2e83**

packages/govern-core/contracts/Govern.sol
```diff
@@ -23,20 +27,24 @@ contract Govern is ERC3000Executor, ACL {
...
             (bool ok, bytes memory ret) = actions[i].to.call{value: actions[i].value}(actions[i].data);
-            require(ok, "govern: call");
-
+            require(ok || allowFailuresMap.get(i), "govern: call");
+            // if a call fails, flip that bit to signal failure
+            failureMap = ok ? failureMap : failureMap.flip(i);
```
kanaagawa, line 41: You can do magic 🎵

packages/govern-core/contracts/pipelines/GovernQueue.sol
```diff
@@ -283,9 +284,9 @@ contract GovernQueue is ERC3000, IArbitrable, ACL {
 
     // Internal
 
-    function _execute(ERC3000Data.Payload memory _payload, bytes32 _containerHash) internal returns (bytes[] memory execResults) {
-        execResults = _payload.executor.exec(_payload.actions);
-        emit Executed(_containerHash, msg.sender, execResults);
+    function _execute(ERC3000Data.Payload memory _payload, bytes32 _containerHash) internal returns (bytes32, bytes[] memory execResults) {
+        emit Executed(_containerHash, msg.sender);
+        return _payload.executor.exec(_payload.actions, _payload.allowFailuresMap, _containerHash);
```
kanaagawa, line 289: Great 🙌

packages/govern-contract-utils/contracts/bitmaps/BitmapLib.sol
```diff
@@ -0,0 +1,19 @@
...
+    bytes32 constant internal empty = bytes32(0);
+    bytes32 constant internal allowAll = empty;
+    bytes32 constant internal denyAll = bytes32(uint256(-1));
+
+    function flip(bytes32 map, uint8 index) internal pure returns (bytes32) {
```
kanaagawa, line 12: Would be great to have an inline comment with an example of the inputs and outputs of each operation.

0xGabi (approved):
I'll include the changes on the subgraph to have the `memo` argument

Evalir:
@izqui Thinking we should get this merged by tomorrow as very very late to get the latest subgraph changes which depend on these. We could do another sanity-check round and merge, and Facu could audit all this later.

PR #71, opened 2020-10-20
Server: add pending review changes

0xGabi:
This PR include the review changes already discussed in #59


packages/govern-server/src/core/utils/network.ts
```diff
@@ -1,13 +1,21 @@
 import { ErrorInvalidNetwork } from '../errors'
-import { NETWORKS } from '../params'
```
nivida, line 2: What do you think about the ``Network`` object I thought to create out of those functions in the first PR?

nivida (commented):
~~Thanks for doing most of my proposed changes! :-)~~
~~It would be interesting to hear if we will do funcDoc's in the future as well? (It would definitely increase the readability of the code; https://jsdoc.app/)~~

-- Found some other stuff and requested changes now for this PR.

packages/govern-server/src/core/data/TheGraphClient.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (typeof options === 'boolean') {
+      console.warn(
+        'TheGraphClient: please use `new TheGraphClient(url, { verbose })` rather than `new TheGraphClient(url, verbose)`.'
+      )
+      options = { verbose: options }
+    }
+    return options
+  }
+
+  async performQuery(
```
nivida: It isn't required to have it on a new line here cause it does not exceed the line-length.

packages/govern-server/src/core/data/TheGraphClient.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  constructor(
+    subgraphUrl: string,
+    options: TheGraphClientOptions | boolean = {}
+  ) {
+    this.#verbose = this.parseOptions(options).verbose ?? false
+    this.#client = new Client({ maskTypename: true, url: subgraphUrl, fetch })
+  }
+
+  private parseOptions(
```
nivida: It isn't required to have it on a new line here cause it does not exceed the line-length.

packages/govern-server/src/core/data/TheGraphClient.ts
```diff
@@ -0,0 +1,71 @@
...
+
+type TheGraphClientOptions = {
+  verbose?: boolean
+}
+
+class TheGraphClient {
+  #client: Client
+  #verbose: boolean
+
+  constructor(
```
nivida: It isn't required to have it on a new line here cause it does not exceed the line-length.

packages/govern-server/src/core/data/index.ts
```diff
@@ -7,53 +7,52 @@ import {
...
       verbose: config.verbose,
     })
   }
 
+  private getSubgraphUrl(network: Network): string | null {
+    if (network.chainId === 1) {
+      return 'placeholder'
+    }
+    if (network.chainId === 4) {
+      return 'https://api.thegraph.com/subgraphs/name/0xgabi/aragon-govern-rinkeby-staging'
```
nivida: Those strings/hosts should probably be in the config file.

packages/govern-server/src/core/config.json
```diff
@@ -0,0 +1,19 @@
+{
+  "networks": {
```
nivida: This should be an array cause it would simplify the code down in the network functions and the names wouldn't be duplicated.

nivida (changes requested):
Thanks for doing most of my proposed changes! :-)

It would be interesting to hear if we will do funcDoc's in the future as well? (It would definitely increase the readability of the code; https://jsdoc.app/)

packages/govern-server/src/core/config.json
nivida: ... or you keep the object and remove the ``name`` property and it would just have to create once an ``array`` out of the ``object`` :-)

packages/govern-server/src/core/utils/network.ts
```diff
@@ -1,84 +1,102 @@
...
-  // We compare with undefined to accept empty strings.
-  if (name === undefined) {
-    name = chainIdNetwork.name
+  get name() {
+    return this.#name
   }
 
-  if (!ensAddress) {
-    ensAddress = chainIdNetwork.ensAddress
+  private networkFromChainId(chainId: number): NetworkType | null {
```
bpierre: Should `networkFromChainId` and `networkFromName` be simple functions or static methods, since they are not really related to `Network` instances?

packages/govern-server/src/core/config.json
```diff
@@ -0,0 +1,21 @@
+{
```
bpierre: Should we define this file as TS, as it is part of the codebase?

packages/govern-server/src/core/config.json
```diff
@@ -0,0 +1,21 @@
...
+      "ensAddress": "0x98df287b6c145399aaa709692c8d308357bc085d"
+    },
+    {
+      "chainId": 100,
+      "name": "xdai",
+      "ensAddress": "0xaafca6b0c89521752e559650206d7c925fd0e530"
+    }
+  ],
+  "subgraphAccount": "0xgabi",
+  "rinkebySubgraphName": "aragon-govern-rinkeby-staging"
```
bpierre: Should this be part of the network configuration? Also should we use `aragon/aragon-govern-rinkeby` for Rinkeby?

packages/govern-server/src/core/errors.ts
```diff
@@ -16,8 +12,12 @@ export class ErrorException extends Error {
...
     // from appearing with the error in the console.
-    defineNonEnumerable(this, 'name', name)
-    defineNonEnumerable(this, 'code', code)
+    this.defineNonEnumerable(this, 'name', name)
+    this.defineNonEnumerable(this, 'code', code)
+  }
+
+  private defineNonEnumerable(instance: object, name: string, value: any) {
+    Object.defineProperty(instance, name, { value, enumerable: false })
   }
```
bpierre: I think we could move this function elsewhere, as it’s pretty generic? But if we want to keep it here, we could use `this` directly rather than passing it as a parameter.

bpierre (approved):
LGTM, I left some comments about minor things but I think we can merge otherwise!

PR #72, opened 2020-10-20
Subgraph: update subgraph with latest smart contract logic

0xGabi:
This PR merge the latests changes from master and update the subgraph to use the latests smart contract changes.

The relevant commit with the subgraph changes is: 2415d0bbec4e1478ac15ca592ed3fc0551e40ecd


packages/govern-subgraph/schema.graphql
```diff
@@ -34,25 +35,25 @@ type Action @entity {
...
-  execution: Execution
+  execution: Execution!
+  queueItem: QueueItem!
 }
 
 type GovernQueue @entity {
   id: ID!
   address: Bytes!
   config: Config
+  queued: [QueueItem!]
```
Evalir: What about changing this to `queuedItems` or `queuedActions`

packages/govern-subgraph/schema.graphql
```diff
@@ -34,25 +35,25 @@ type Action @entity {
...
+  challenges: [Challenge!]
+  vetos: [Veto!]
+  executions: [Execution!] @derivedFrom(field: "queue")
```
Evalir: Hmmm, I think we need to further see what's the relation between a `QueueItem` and an execution. For me, a `QueueItem` becomes an `Execution` after the executor has called its `exec` function. Mostly thinking about duplicated entities.

Evalir (commented):
Will review more thoroughly later:

izqui:
this can now be rebased/merged against master

packages/govern-subgraph/schema.graphql
kanaagawa: You are right, I think there are room for optimize the number of entities.

packages/govern-subgraph/schema.graphql
```diff
@@ -8,46 +8,39 @@ type ERC3000Registry @entity {
...
   config: Config
-  queued: [QueueItem!]
-  challenges: [Challenge!]
-  vetos: [Veto!]
-  executions: [Execution!] @derivedFrom(field: "queue")
   games: [OptimisticGame!] @derivedFrom(field: "queue")
+  scheduledPackets: [Schedule!]
+  challengedPackets: [Challenge!]
+  vetoedPackets: [Veto!]
+  executedPackets: [Execution!] @derivedFrom(field: "queue")
```
bpierre, line 40: I find it a bit odd that these are named `somethingPackets`, but point to wrappers (types with a `.packet`) rather than actual packets (as in `QueuePacket`). Should we just stick to `scheduled`, `challenged`, etc. to better reflect the types?

Evalir: Also proposed this on #82; I think it would make sense to simplify the schema further to be closer to actual types sent by the contract.

kanaagawa: Make sense, I was trying to reflect that the `packets` are the things being handle. But depending on the type you also have extra information related to that action.

PR #76, opened 2020-10-20
Console: update query to new subgraph

0xGabi:
This PR update the console to use the latest version of the subgraph.

A requirement is to have a `.env` file with the following variable: 
```
REACT_APP_SUBGRAPH_URL=https://api.thegraph.com/subgraphs/name/aragon/aragon-govern-rinkeby
```


Evalir (approved):
LGTM! Could you give me the new subgraph URL so I can properly set it up in fleek?

0xGabi:
Yes, the current subgraph URL is: `https://api.thegraph.com/subgraphs/name/aragon/aragon-govern-rinkeby`

Note that we are going to deploy to our own infra and that will change.

PR #77, opened 2020-10-21
Govern JS Client

### LINKED ISSUE #75: Govern.js
- JS library (browser + node) fetching data from a Govern Server, through its GraphQL API.
- Reflects the [Govern Server JS API](https://github.com/aragon/govern/blob/a14370b6bd248ecd3b956815113dd980c1391484/packages/govern-server/src/core/data/index.ts#L76-L131). 
- Inspired by the [Connect](https://connect.aragon.org/) API.
- Actions resulting in transactions can go through a local Ethereum provider or the Govern Server.

Should we use `@aragon/govern.js` as a package name? @izqui 

Implementation: https://github.com/aragon/govern/pull/77

nivida:
This PR implements the current Govern Server API.

**I have switched it to ready for review but the API queries need to be adjusted and tested as soon as the Govern server API has updated**

closes #75


.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules
 .vscode
 .env
-.DS_Store
\ No newline at end of file
+.DS_Store
+.idea/
```
bpierre: I think we should remove these (and `.vscode` too): they belong to local / OS configuration (`~/.gitignore`) and are not related to the repository.

packages/govern/README.md
```diff
@@ -0,0 +1,32 @@
...
+## Usage
+
+**Usage with default config**
+
+> The default config uses the Govern servers deployed by Aragon
+
+``` javascript
+import { dao } from '@aragon/govern'
+
+const response = await dao('0x0...');
```
bpierre: I’m thinking now that having the a functional API will make it impossible to do `const dao = …` since `dao` is already taken by the function name. But maybe it’s not that bad, since it will force people to name what they fetch? e.g. `const myDao = await dao()`

packages/govern/internal/actions/DAOAction.ts
```diff
@@ -0,0 +1,45 @@
+import { DocumentNode } from 'graphql'
+import AbstractAction from './lib/AbstractAction'
+import dao from '../clients/graphql/queries/dao'
+import { isAddress } from '@ethersproject/address'
+
+/**
+ * @class DAOAction
+ */
+export default class DAOAction extends AbstractAction {
```
bpierre:
```suggestion
export default class DaoAction extends AbstractAction {
```

:smile:

packages/govern/internal/actions/DAOSAction.ts
```diff
@@ -0,0 +1,17 @@
+import { DocumentNode } from 'graphql'
+import AbstractAction from './lib/AbstractAction'
+import daos from '../clients/graphql/queries/daos'
+
+/**
+ * @class DAOSAction
+ */
+export default class DAOSAction extends AbstractAction {
```
bpierre:
```suggestion
export default class DaosAction extends AbstractAction {
```

packages/govern/internal/actions/lib/AbstractAction.ts
```diff
@@ -0,0 +1,72 @@
...
+   * Validates the given parameters
+   *
+   * @method validateParameters
+   *
+   * @param {any} parameters
+   *
+   * @returns {any}
+   */
+  protected validateParameters(parameters?: any): any {
+    return parameters;
```
bpierre: We might want to run prettier on all the files:

```suggestion
    return parameters
```

packages/govern/internal/actions/QueuesForDAOAction.ts
```diff
@@ -0,0 +1,44 @@
...
+   * @protected
+   */
+  protected gqlQuery: DocumentNode = queuesForDAO
+
+  /**
+   * @param {Object} parameters
+   *
+   * @constructor
+   */
+  constructor(parameters: { address: string }) {
```
bpierre: Should we have a `type Address = string` like in the Govern Server package?

packages/govern/internal/clients/graphql/GraphQLClient.ts
```diff
@@ -0,0 +1,103 @@
...
+ */
+export default class GraphQLClient implements ClientInterface {
+  /**
+   * The urql/core Client object
+   *
+   * @var {Client} client
+   *
+   * @private
+   */
+  private client: Client
```
bpierre, line 19: Should we rename `Client` into `UrqlClient` for clarity?

```suggestion
  private urlqClient: UrqlClient
```

packages/govern/internal/clients/graphql/GraphQLClient.ts
```diff
@@ -0,0 +1,103 @@
...
+   * @method request
+   *
+   * @param {DocumentNode} query
+   * @param {Object} args
+   *
+   * @returns {Promise<any>}
+   *
+   * @public
+   */
+  public async request(query: DocumentNode, args: any = {}): Promise<any> {
```
bpierre: I think we could import and use [`OperationResult`](https://github.com/aragon/govern/blob/graphql-api/packages/govern-server/src/core/types.ts#L1) to type the returned value here.

packages/govern/internal/clients/graphql/GraphQLClient.ts
```diff
@@ -0,0 +1,103 @@
...
+   *
+   * @method mapError
+   *
+   * @param {any} error
+   *
+   * @returns {string}
+   *
+   * @private
+   */
+  private mapError(error: any): string {
```
bpierre:
```suggestion
  private mapError(error: Error): string {
```

bpierre (commented):
Looking really good! I really like the separation you make between public API and internal implementation.

What would you think of trying to replace some `any` by direct types when possible (e.g. by importing them from urql), or by using generics / `is` / `unknown`?

packages/govern/README.md
nivida: This isn't that bad.. they have also the possibility to rename it with the import :)

packages/govern/internal/actions/QueuesForDAOAction.ts
nivida: Will do that 👌

packages/govern/internal/clients/graphql/GraphQLClient.ts
nivida: I think the comment brings up enough clarity :)

packages/govern/internal/actions/DAOSAction.ts
nivida: I will do that.. that's why asked for the Aragon Code Style document.

nivida:
I will set the PR to ready-for-review as soon as I did a review on my own ;-) @bpierre

docs/developers/govern-js.md
```diff
@@ -0,0 +1,84 @@
+# Govern.js API
```
bpierre, line 1: I think we should add this document to `docs/SUMMARY.md`.

packages/govern/README.md
```diff
@@ -0,0 +1,90 @@
+# govern.js
+
+> The JavaScript client API for the Govern server
+
+## Usage
```
bpierre: Should we just point to the documentation website?

bpierre (approved):
LGTM (G as in great)

packages/govern/README.md
nivida: Sounds good!

PR #80, opened 2020-10-21
erc3k: update interface id

0xGabi:
Update the `ERC3000_INTERFACE_ID` after the changes to the `IERC3000`

### DIFF
--- a/packages/govern-create/buidler.config.js
+++ b/packages/govern-create/buidler.config.js
@@ -44,13 +44,20 @@ task('deploy-registry', 'Deploys an ERC3000Registry instance').setAction(
 
 task('deploy-factory', 'Deploys an GovernBaseFactory instance').setAction(
   async (_, { ethers }) => {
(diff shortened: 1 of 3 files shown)


packages/govern-create/buidler.config.js
```diff
@@ -90,7 +97,7 @@ task('deploy-govern', 'Deploys an Govern from provided factory')
       factoryAddr
     )
     const tx = await GovernBaseFactory.newDummyGovern(name, {
-      gasLimit: 3500000,
+      gasLimit: 4000000,
```
Evalir, line 100: Interesting; did the gas requirements for deploying a DAO go up?

Evalir (approved):
🔥🍾💥 let's get this in!

packages/govern-create/buidler.config.js
kanaagawa: This is for deploying the Govern Factory so we are fine. But with the previous value was reverted

PR #82, opened 2020-10-22
Console: update query with new subgraph changes + use for to iterate action on subgraph

0xGabi:


packages/govern-console/src/pages/DaoView.tsx
```diff
@@ -42,19 +42,21 @@ const DAO_QUERY = gql`
...
         }
-        executions {
+        executedPackets {
           id
           queue {
             address
           }
         }
-        queued {
+        scheduledPackets {
```
Evalir: Tbh I'm not a fan of these names... I'd just use `Containers` and `executedContainers`, but even that seems a bit redundant as a `Container` should have an `executionStatus`.

Evalir (commented):
Some comments:

PR #86, opened 2020-10-23
Subgraph: rename to align with contracts

0xGabi:
This PR rename the current schema to align more with the smart contracts. Apologies if I introduce further confusion with the naming. The other current options we discussed are: `Container`, `Packet` and `Item`.

Let's use this branch to stick to one of the above.


packages/govern-subgraph/schema.graphql
```diff
@@ -63,37 +63,37 @@ type ERC20 @entity {
   address: Bytes!
 }
 
-type QueuePacket @entity {
+type Payload @entity {
   id: ID!
   queue: GovernQueue!
-  status: PacketStatus!
+  status: Status!
```
bpierre: Should we use the term “state” to reflect the contracts?

Also it seems that the hierarchy is the following in the contracts (using TS to describe the relations):

```ts
type Queue = {
  initConfig: Config

  // In the contract only the state is stored in the queue using
  // the container hash as a key, but the subgraph could be more
  // straightforward and have containers in the queue directly.
  queued: Container[]
}

type Config = {
  executionDelay: BigInt
  scheduleDeposit: Collateral
  challengeDeposit: Collateral
  vetoDeposit: Collateral
  resolver: string
  rules: string
}

type Collateral = {
  address: string
  amount: BigInt
}

type Container = {
  config: Config

  // I don’t think it would make sense for a Container to belong to different
  // queues, so I think we can store the state on the Container directly.
  state: ContainerState

  payload: ContainerPayload
}

enum ContainerState {
  None,
  Scheduled,
  Challenged,
  Approved,
  Rejected,
  Cancelled,
  Executed
}

type ContainerPayload = {
  nonce: BigInt
  executionTime: BigInt
  submitter: string
  executor: Executor
  actions: Action[]
  allowFailuresMap: string
  proof: string
}

type Action = {
  to: string
  value: BigInt
  data: string
}
```

Should we try to reproduce this in the subgraph?

Note: I excluded having references to parents for readability, but in the subgraph we could also expose them (e.g. by having `Action.container`).

packages/govern-subgraph/schema.graphql
```diff
@@ -130,7 +130,7 @@ type Role @entity {
   frozen: Boolean!
 }
 
-enum PacketStatus {
+enum Status {
```
bpierre: Does it make sense to have `Vetoed` as a status? Should we instead set it to `Cancelled` [like in the contract](https://github.com/aragon/govern/blob/master/packages/govern-core/contracts/pipelines/GovernQueue.sol#L203) and, to store the different events, have a `Container.history = ContainerEvent[]` that would contain the entire history of events happening on a given container?

packages/govern-subgraph/schema.graphql
Evalir: 100% agree with this! This is quite similar to the first version of the subgraph, and I think having a structure closer to the contracts lets us understand the subgraph faster.

> Note: I excluded having references to parents for readability, but in the subgraph we could also expose them (e.g. by having Action.container).

Yep, I agree having the parent references in the schema.

kanaagawa: This structure looks really clean and super close to the smart contract. I diverge a bit from that with the intention to make it a bit more straight forward. And probably has the opposite effect 🙈

I think I miss-understood the importance of having a `Container` that have both a reference to the `Payload` and the `Config`. Because there are scenarios where the `Config` changed and you still want to have the `Payload` linked to the previous `Config`.

packages/govern-subgraph/schema.graphql
kanaagawa: Yes, that's a good point. I think we previously had a `Vetoed` state but I agree we should stick to the contracts. I like the idea of having a `Container` history 🙌

0xGabi:
@bpierre @Evalir I just include the changes discussed. Would you mind take a second look?

Pierre, about having a `Container.history` with all the history of events. What information you think is relevant to store?

bpierre:
> Pierre, about having a `Container.history` with all the history of events. What information you think is relevant to store?

I think we could have the date of the event, its type, and some data?

The event types would slightly differ from `Container.state`, we could use the present tense to make it clear that they are not the same thing:

```ts
enum ContainerEventType {
  Schedule,
  Challenge,
  Approve,
  Reject,
  Cancel,
  Execute,
  Veto,
  Rule,
  EvidenceSubmission
}

type ContainerEvent = {
  type: ContainerEventType
  time: BigInt
  data: string[] // arbitrary data related to the event
}
```

packages/govern-subgraph/schema.graphql
```diff
@@ -24,15 +24,15 @@ type Govern @entity {
...
+  container: Container!
   results: [Bytes!]
   createdAt: BigInt!
 }
 
 type GovernQueue @entity {
   id: ID!
   address: Bytes!
-  config: Config
+  configs: [Config!]
```
Evalir: I'd prefer to only keep a single config here—We could totally choose to track the history later, but let's keep it close to the contracts.

packages/govern-subgraph/schema.graphql
```diff
@@ -78,22 +80,22 @@ type Payload @entity {
...
   data: Bytes!
 }
 
 type Schedule @entity {
   id: ID!
-  payload: Payload!
+  container: Container!
   collateral: Collateral!
   createdAt: BigInt!
 }
```
Evalir: This entity, now that we have containers, feels a bit redundant.

Evalir (commented):
Some more comments, I'm probably gonna do these soon:

packages/govern-subgraph/schema.graphql
```diff
@@ -16,28 +16,18 @@ type Govern @entity {
   id: ID!
   address: Bytes!
   metadata: Bytes
-  games: [OptimisticGame!] @derivedFrom(field: "executor")
-  executions: [Execution!]
+  correspondingEntries: [RegistryEntry!] @derivedFrom(field: "executor")
```
Evalir: Still not happy with this name, but going with it for now. Suggestions are welcome. :D

packages/govern-subgraph/schema.graphql
```diff
@@ -86,13 +84,8 @@ type Action @entity {
...
-type Schedule @entity {
-  id: ID!
-  container: Container!
-  collateral: Collateral!
-  createdAt: BigInt!
-}
-
+# I honestly think this should not be here,
+# but we can handle this later.
 type Challenge @entity {
```
Evalir: As I say here, I really think that this entity should not exist here. It seems out of scope for the subgraph itself because if an action gets disputed, it's really up to the dispute resolution protocol to be able to get this info from its own graph; we should only facilitate finding the actual disputed action so that they're able to show context regarding it.

Removing this required some hefty changes from the subgraph, so I avoided it, but definitely I think at most, we should just set it to the actual `Challenged` state, and on the `history` field provide some more info on the `data` field.

packages/govern-subgraph/schema.graphql
```diff
@@ -114,15 +107,6 @@ type Evidence @entity {
   createdAt: BigInt!
 }
 
-type Veto @entity {
```
Evalir, line 115: For removing this, on the `data` field of the latest container `history` event that will get recorded I just put the `reason`. That way you should be able to see why this was vetoed.

Evalir (commented):
Changed some more things, mainly to align with @bpierre's suggestions. This is not 100% there yet, but we're very close! The only thing left to be in sync with the server, should be removing the `Challenge` entity.

Evalir:
Also, [a fix is coming up](https://github.com/graphprotocol/ethabi/pull/15) for the issue we're having with the `Executed` event not being found in the contracts, so as soon as we merge this we can just test this heavily with some scripts I've been making 😄

Evalir:
cc @0xGabi I'm shooting to have this merged by 4PM UTC tomorrow—by that time The Graph's fix should be up and we can inmediately do a smoke test on it. :)

packages/govern-subgraph/schema.graphql
bpierre: If we go for the name `RegistryEntry` to refer to the Dao <> Queue association, should we keep it consistent and use `registryEntries` here?

Or maybe we could use the term `Registration` (`registrations` here) to reflect [the verb used for the method](https://github.com/aragon/govern/blob/master/packages/erc3k/contracts/ERC3000Registry.sol#L18)? We could also do `DaoQueueRegistration` but it starts to be a bit long :smile:

Also I don’t think the array can be null?

```suggestion
  correspondingEntries: [RegistryEntry!]! @derivedFrom(field: "executor")
```

packages/govern-subgraph/schema.graphql
```diff
@@ -52,48 +42,53 @@ type Config @entity {
...
+  actions: [Action!] @derivedFrom(field: "payload")
   allowFailuresMap: Bytes!
   proof: Bytes!
 }
 
+type ContainerEvent @entity {
+  id: ID!
+  type: ContainerEventType
+  createdAt: BigInt
+  data: [String]!
```
bpierre: Wondering if we should use `[Bytes!]!` here, to accept any type of data?

packages/govern-subgraph/schema.graphql
```diff
@@ -130,13 +116,26 @@ type Role @entity {
...
+enum ContainerEventType {
+  Schedule,
+  Challenge,
+  Approve,
+  Reject,
+  Cancel,
+  Execute,
+  Veto,
+  Rule,
+  EvidenceSubmission
```
bpierre: Maybe we should rename this `SubmitEvidence` to make it a verb like the others?

packages/govern-subgraph/src/GovernQueue.ts
```diff
@@ -11,148 +11,182 @@ import {
...
 
 export function handleExecuted(event: ExecutedEvent): void {
-  const packet = loadOrCreateQueuePacket(event.params.containerHash, event)
-
-  packet.status = EXECUTED_STATUS
+  const container = loadOrCreateContainer(event.params.containerHash, event)
+  container.state = EXECUTED_STATUS
 
-  packet.save()
+  container.save()
```
bpierre, line 81: Should we create a `containerEvent.type = EXECUTE_CONTAINER_EVENT` here?

Also should we have container events for:

- `handleResolved`
- `handleEvidenceSubmitted`
- `handleRuled`

And remove the ones we are not using from the list?

packages/govern-subgraph/schema.graphql
kanaagawa: I think games were pretty appealing given that DAOs are going to be participating on this optimistic games. What is the reason to instead linked to entities on a registry?

packages/govern-subgraph/schema.graphql
```diff
@@ -16,28 +16,18 @@ type Govern @entity {
...
   address: Bytes!
   metadata: Bytes
-  games: [OptimisticGame!] @derivedFrom(field: "executor")
-  executions: [Execution!]
+  correspondingEntries: [RegistryEntry!] @derivedFrom(field: "executor")
+  containers: [Container]!
   roles: [Role!]
 }
 
-type Execution @entity {
```
kanaagawa, line 24: I think we need to have some way to track the execution that a Govern instance does so we keep a record of the results of the given execution.

packages/govern-subgraph/schema.graphql
Evalir: DAOs might be participating in optimistic games, but they're _not_ optimistic games by themselves. I think it's easier to think of these entities as entries in the registry, mainly because it's what they are in the contracts.

packages/govern-subgraph/schema.graphql
Evalir: We could probably just record these on the `data` field of the corresponding `ContainerEvent`?

packages/govern-subgraph/schema.graphql
```diff
@@ -16,28 +16,18 @@ type Govern @entity {
   id: ID!
   address: Bytes!
   metadata: Bytes
-  games: [OptimisticGame!] @derivedFrom(field: "executor")
-  executions: [Execution!]
+  correspondingEntries: [RegistryEntry!] @derivedFrom(field: "executor")
+  containers: [Container]!
```
bpierre, line 20:
```suggestion
  containers: [Container!]!
```

packages/govern-subgraph/src/Govern.ts
```diff
@@ -7,23 +7,17 @@ import {
...
 } from '../generated/schema'
 import { frozenRoles, roleGranted, roleRevoked } from './lib/MiniACL'
 
 export function handleExecuted(event: ExecutedEvent): void {
   const govern = loadOrCreateGovern(event.address)
 
-  const execution = loadOrCreateExecution(event)
+  const currentContainers = govern.containers
 
-  execution.results = event.params.execResults
```
kanaagawa, line 19: This is the only place we can keep track of the results of the execution. Maybe have it as part of the `Execute` container event.

packages/govern-subgraph/schema.graphql
```diff
@@ -16,28 +16,18 @@ type Govern @entity {
...
-  config: Config
-  games: [OptimisticGame!] @derivedFrom(field: "queue")
-  scheduled: [Schedule!]
-  challenged: [Challenge!]
-  vetoed: [Veto!]
-  executed: [Execution!] @derivedFrom(field: "queue")
+  config: Config!
+  correspondingEntries: [RegistryEntry!] @derivedFrom(field: "queue")
+  queued: [Container!]
+  challenged: [Challenge]!
```
bpierre: I might be wrong, but I think we can use `[E!]!` most of the time (non-null arrays of non-null entries)?

```suggestion
  correspondingEntries: [RegistryEntry!]! @derivedFrom(field: "queue")
  queued: [Container!]!
  challenged: [Challenge!]!
```

packages/govern-subgraph/src/Govern.ts
Evalir: Yeah, definitely agree. :)

packages/govern-subgraph/schema.graphql
kanaagawa: I choose to keep track the history of configs because we need to know the specific configuration that a `Container` was schedule with. Otherwise when we update the config of the `Queue` the `Container` will have a the wrong configuration data.

packages/govern-subgraph/src/GovernQueue.ts
```diff
@@ -11,148 +11,182 @@ import {
...
-  packet.nonce = event.params.payload.nonce
-  packet.executionTime = event.params.payload.executionTime
-  packet.submitter = event.params.payload.submitter
-  packet.executor = event.params.payload.executor.toHexString()
-  packet.allowFailuresMap = event.params.payload.allowFailuresMap
-  packet.proof = event.params.payload.proof
-
-  const schedule = new ScheduleEntity(event.params.containerHash.toHexString())
-
-  const scheduleDeposit = loadOrCreateCollateral(event, '1')
```
kanaagawa, line 58: We need to track the `Collateral` at the time the `Container` was schedule.

packages/govern-subgraph/schema.graphql
Evalir: Btw @bpierre Yep, def agree about the naming change. What about sticking to `registryEntry` and `registryEntries`? I think it sounds natural, even if it's 100% aligned with the method's verb. :)

packages/govern-subgraph/src/GovernQueue.ts
```diff
@@ -11,148 +11,182 @@ import {
...
+    event.params.containerHash,
+    amountOfEvents
+  )
+  const containerEventId = buildActionId(
+    event.params.containerHash,
+    amountOfEvents
+  )
+  containerEvent.createdAt = event.block.timestamp
+  containerEvent.type = SCHEDULE_CONTAINER_EVENT
+  containerEvent.data = []
```
kanaagawa: Maybe we can include the collateral as part of the data here.

packages/govern-subgraph/schema.graphql
Evalir: Yep yep, I was meaning to change this but that slipped by. 😄

packages/govern-subgraph/schema.graphql
bpierre: Sounds good to me! @0xGabi what do you think?

kanaagawa: Yes, sound good

packages/govern-subgraph/src/Govern.ts
```diff
@@ -5,17 +5,25 @@ import {
...
 export function handleExecuted(event: ExecutedEvent): void {
   const govern = loadOrCreateGovern(event.address)
+  const container = loadOrCreateContainer(event.params.memo)
 
-  const currentContainers = govern.containers
+  handleContainerEvent(
+    container,
+    event.block.timestamp,
+    EXECUTE_CONTAINER_EVENT,
+    [event.params.execResults.toString()]
```
kanaagawa: I was not sure if we should instead return a data element per result here.

packages/govern-subgraph/schema.graphql
```diff
@@ -67,7 +67,7 @@ type ContainerEvent @entity {
   id: ID!
   type: ContainerEventType
   createdAt: BigInt
-  data: [String]!
+  data: [String!]!
```
Evalir: Btw, small thing, let's change this to bytes so we can encode anything here!

kanaagawa: The main reason I didn't do it yet was because I'm not sure how to handle `BigInt`. What encoding do you suggest?

PR #87, opened 2020-10-25
Core, utils: Implement adaptative ERC165

izqui:


packages/erc3k/contracts/ERC3000.sol
```diff
@@ -5,21 +5,66 @@
...
+          ^ this.execute.selector
+          ^ this.challenge.selector
+          ^ this.resolve.selector
+          ^ this.veto.selector
+          ^ this.configure.selector
+        ;
+
+    /**
+     * @notice Schedules an action for execution, allowing for challenges and vetos on a defined time window
+     * @param container A Container struct holding both the paylaod being scheduled for execution and
```
Evalir:
```suggestion
     * @param container A Container struct holding both the payload being scheduled for execution and
```

packages/erc3k/contracts/ERC3000.sol
```diff
@@ -5,21 +5,66 @@
...
+       the current configuration of the system
+     */
+    function schedule(ERC3000Data.Container memory container) virtual public returns (bytes32 containerHash);
+    event Scheduled(bytes32 indexed containerHash, ERC3000Data.Payload payload, ERC3000Data.Collateral collateral);
+
+    /**
+     * @notice Executes an action after its execution delayed has passed and its state hasn't been altered by a challenge or veto
```
Evalir:
```suggestion
     * @notice Executes an action after its execution delay has passed and its state hasn't been altered by a challenge or veto
```

packages/erc3k/contracts/ERC3000.sol
```diff
@@ -5,21 +5,66 @@
...
+     * @param container A ERC3000Data.Container struct holding both the paylaod being scheduled for execution and
+       the current configuration of the system
+     * should be a MUST payload.executor.exec(payload.actions)
```
Evalir:
```suggestion
     * MUST be an ERC3000Executor call: payload.executor.exec(payload.actions)
```

packages/erc3k/contracts/ERC3000.sol
```diff
@@ -5,21 +5,66 @@
...
+     */
+    function execute(ERC3000Data.Container memory container) virtual public returns (bytes32 failureMap, bytes[] memory execResults);
+    event Executed(bytes32 indexed containerHash, address indexed actor);
+
+    /**
+     * @notice Challenge a container in case its scheduling is illegal as per Config.rules. Pulls collateral and dispute fees from sender into contract
+     * @param container A ERC3000Data.Container struct holding both the paylaod being scheduled for execution and
```
Evalir:
```suggestion
     * @param container A ERC3000Data.Container struct holding both the payload being scheduled for execution and
```

packages/erc3k/contracts/ERC3000.sol
```diff
@@ -5,21 +5,66 @@
...
+       the current configuration of the system
+     * @param reason Hint for case reviewers as to why the scheduled container is illegal
```
Evalir, line 42: Maybe we could expand this to make it clear that you could, for example, submit an IPFS CID for the dispute resolution system to pick up and use as the challenger's argument.

packages/erc3k/contracts/ERC3000.sol
```diff
@@ -5,21 +5,66 @@
...
+     */
+    function challenge(ERC3000Data.Container memory container, bytes memory reason) virtual public returns (uint256 resolverId);
+    event Challenged(bytes32 indexed containerHash, address indexed actor, bytes reason, uint256 resolverId, ERC3000Data.Collateral collateral);
+
+    /**
+     * @notice Apply arbitrator's ruling over a challenge once it has come to a final ruling
+     * @param container A ERC3000Data.Container struct holding both the paylaod being scheduled for execution and
```
Evalir:
```suggestion
     * @param container A ERC3000Data.Container struct holding both the payload being scheduled for execution and
```

packages/govern-contract-utils/contracts/adaptative-erc165/AdaptativeERC165.sol
```diff
@@ -0,0 +1,55 @@
+/*
+ * SPDX-License-Identifier:    MIT
+ */
+
+pragma solidity ^0.6.8;
+
+import "../erc165/ERC165.sol";
+
+contract AdaptativeERC165 is ERC165 {
```
Evalir, line 9: Did we test gas performance for this contract? I haven't, but we should make sure `suppportsInterface` spends less than 30k gas as per [the spec](https://eips.ethereum.org/EIPS/eip-165).

Evalir (commented):
Some comments, mostly fixing spelling.

When are we looking to merge this? This is a breaking change (again hehe) for contracts and we'll need to re-deploy / adapt most, if not all of the periphery packages here. Ideally this would be merged tomorrow.

packages/govern-contract-utils/contracts/adaptative-erc165/AdaptativeERC165.sol
```diff
@@ -0,0 +1,55 @@
...
+        bytes32 magicNumber = callbackMagicNumbers[_sig];
+        require(magicNumber != UNREGISTERED_CALLBACK, "adap-erc165: unknown callback");
+
+        emit ReceivedCallback(_sig, _data);
+
+        // low-level return magic number
+        assembly {
+            mstore(0x00, magicNumber)
+            return(0x00, 0x20)
+        }
```
kanaagawa, line 35: Why we need to use assembly here?

packages/govern-core/contracts/Govern.sol
```diff
@@ -5,28 +5,36 @@
...
+        _registerStandard(ERC3000_EXEC_INTERFACE_ID);
     }
 
     receive () external payable {
         emit ETHDeposited(msg.sender, msg.value);
     }
 
+    fallback () external {
+        _handleCallback(msg.sig, msg.data);
+    }
```
kanaagawa, line 36: I'm not sure I fully understand the propose of `handleCallback` as the fallback method for `Govern`.

packages/govern-core/test/pipelines/govern-queue.test.ts
```diff
@@ -34,6 +35,25 @@ describe('Govern Queue', function () {
...
     )) as GovernQueueFactory
-    gq = await GQ.deploy(ownerAddr, INIT_CONFIG)
+    gq = await GQ.deploy(ownerAddr, INIT_CONFIG) as GovernQueue
+  })
+
+  context('ERC-165', () => {
+    const ERC165_INTERFACE_ID = '0x01ffc9a7'
+    const ERC3000_INTERFACE_ID = '0x6d74ef88'
+
+    // TODO: Fix types @gabi
```
kanaagawa: I review and ran the tests and the types are fine to me. What you mean?

0xGabi (approved):
The changes looks good to me. I think the Adaptative ERC165 is a great addition to Optimistic DAOs to keep them flexible.

I left a few comments to better understand how the callback mechanism work. It's the part I'm a bit confuse still.

packages/govern-contract-utils/contracts/adaptative-erc165/AdaptativeERC165.sol
izqui: good point. this really only does 1 SLOAD which is under 1k gas rn

packages/govern-contract-utils/contracts/adaptative-erc165/AdaptativeERC165.sol
izqui: Because we need to do a low-level return (have a fallback return stuff)

packages/govern-core/contracts/Govern.sol
izqui: We are basically allowing to install new `onXXX` (e.g. ERC721, 1155) callbacks into Govern at runtime. We catch certain function sigs in the callback, log the data it has and then return a magic number

packages/govern-core/test/pipelines/govern-queue.test.ts
izqui: I fixed it by adding the `expect` import that was missing and ts was giving me some weird issues

izqui:
```suggestion
```

packages/govern-core/test/pipelines/govern-queue.test.ts
```diff
@@ -34,6 +35,24 @@ describe('Govern Queue', function () {
...
     )) as GovernQueueFactory
-    gq = await GQ.deploy(ownerAddr, INIT_CONFIG)
+    gq = await GQ.deploy(ownerAddr, INIT_CONFIG) as GovernQueue
+  })
+
+  context('ERC-165', () => {
+    const ERC165_INTERFACE_ID = '0x01ffc9a7'
+    const ERC3000_INTERFACE_ID = '0x6d74ef88'
+
+
```
izqui:
```suggestion

```

packages/govern-core/test/pipelines/govern-queue.test.ts
kanaagawa: Oh right, with the addition of typescript buidler required to be explicit about the imports.

packages/govern-core/contracts/Govern.sol
kanaagawa: Wow this is done at runtime! Maybe I'm confusing the subjects here but the logic for handling those new standards is happening off chain and we check the return magic number is correct? If that's the case this new paradigm of stateless Ethereum is really impressive.

PR #90, opened 2020-10-27
Cleanup: core & create packages

Evalir:
Tidies up a few files in the `govern-core` and `govern-create` packages.


izqui:
For future reference (and we should have a style guide), the commit name should have been `Core, create: cleanup`

PR #91, opened 2020-10-27
Add Govern Token

izqui:
TODO:

- [ ] Copy and adapt tests from Uniswap MerkleDistributor
- [ ] Adapt govern-token tests to general repo style


packages/govern-token/contracts/GovernToken.sol
```diff
@@ -0,0 +1,9 @@
+/*
+ * SPDX-License-Identifier:    MIT
+ */
+
+pragma solidity 0.6.8;
+
+contract GovernToken {
+
```
Evalir: #NoBugs

PR #101, opened 2020-10-28
Subgraph container events: use a shared interface

bpierre:
Add a `ContainerEvent` interface implemented by every container event entity.

Also:

- Add a `finished` property to the `SubmitEvidence` container event.
- Use a separate handler for each container event type.


nivida (approved):
LGTM! 💪

packages/govern-subgraph/schema.graphql
```diff
@@ -111,3 +94,57 @@ enum ContainerState {
   Cancelled
   Executed
 }
+
+# Container events
```
kanaagawa, line 97: Solid 😎

packages/govern-subgraph/src/ERC3000Registry.ts
```diff
@@ -14,16 +14,16 @@ import {
 import { loadOrCreateGovern } from './Govern'
 
 export function handleRegistered(event: RegisteredEvent): void {
-  const registry = loadOrCreateRegistry(event.address)
+  let registry = loadOrCreateRegistry(event.address)
```
kanaagawa, line 17: Why we choose to use `let` instead of `const`?

packages/govern-subgraph/src/utils/events.ts
```diff
@@ -1,42 +1,164 @@
...
+} from '../../generated/templates/GovernQueue/GovernQueue'
+import { buildId, buildIndexedId } from './ids'
+
+function finalizeContainerEvent<T, U>(
+  container: ContainerEntity,
+  containerEvent: U,
+  ethereumEvent: T
+): U {
+  containerEvent.createdAt = ethereumEvent.block.timestamp
+  container.history.push(containerEvent.id)
```
kanaagawa: I think if we only push the new entity here the `Container.hisotry` is not updated on the graph-node. Instead we should:
```
	let containerEventsHistory = container.history
	containerEventsHistory(containerEvent.id)
	container.history = containerEventsHistory
```

packages/govern-subgraph/src/GovernQueue.ts
```diff
@@ -82,79 +81,64 @@ export function handleScheduled(event: ScheduledEvent): void {
...
-      event.params.reason.toHex(),
-      ConfigEntity.load(queue.config).resolver.toHex(),
-      event.params.resolverId.toHex(),
-      event.params.collateral.token.toHex(),
-      event.params.collateral.amount.toHex()
-    ]
-  )
+  let containerEvent = handleContainerEventChallenge(container, event)
+
+  // TODO
```
kanaagawa: Why we left this as a TODO?

0xGabi (changes requested):
Looking really good. The container event abstraction is great!

Left a few request for changes.

packages/govern-subgraph/src/utils/events.ts
kanaagawa: Was trying to find the piece of documentation where they explain this but I couldn't find it.

packages/govern-subgraph/src/ERC3000Registry.ts
nivida: PR #102 explains why :-)

packages/govern-subgraph/src/GovernQueue.ts
bpierre: oops completely forgot this − I need to check something with you about it!

packages/govern-subgraph/src/utils/events.ts
bpierre: Oooh that’s why. I wonder if we could have a utility function to make the intention clear? Will look into it.

Evalir: Heheh yeah this is a bit tricky... I honestly wonder if this is a Assemblyscript limitation or just something that's been overlooked. 😅

Evalir (approved):
Apart from the pending todo and creating a new array instead of directly pushing to the entity thing, this LGTM! I'm soo looking forward to getting this merged.

0xGabi (approved):
LGTM

packages/govern-subgraph/src/utils/events.ts
bpierre: I was reading about it, it is due to the type conversion happening in the schema:



This is how arrays get handled in the generated code, so accessing them in any way creates a copy rather than returning a reference.

kanaagawa: Thanks a lot for sharing!

PR #102, opened 2020-10-28
Subgraph: fix build warnings

bpierre:
AssemblyScript converts `const` expressions into Wasm constants at compile time, which excludes using mutable values. It seems to be smart enough to convert them into variables and display a warning, but using `let` has the same effect without the warnings.


nivida (approved):

Evalir (approved):
I didn't know this 😄 LGTM!

0xGabi:
Good catch! 💪

PR #110, opened 2020-10-29
Subgraph docs: add instructions to deploy locally

bpierre:


packages/govern-subgraph/README.md
```diff
@@ -4,10 +4,85 @@
...
+```
+
+Replacing:
+
+- `<GRAPH_KEY>` by your The Graph key (this is only needed when deploying on The Graph).
+- `<THEGRAPH_USERNAME>` by the username used for your subgraph (usually your GitHub username).
+- `<SUBGRAPH_NAME>` by the name of the subgraph itself.
+- `<DESIRED_NETWORK>` by one of the available networks (see package.json).
+
+### Deploy the subgraph locally
```
nivida, line 47: I think this section isn't required to be that long cause of the DevEnv I've added. Also commands to deploy it locally already exist.

bpierre: Yes, I was hesitating between adding this section or just telling people to run the server dev env, but I thought we should make it evolve a bit before, by:

- Moving it to its own package (decoupling it from the server).
- Having the possibility to change the Ethereum RPC endpoint.
- Having the possibility to run it without the server (i.e. a command that only runs the docker-compose).

Then it would be great to also have the possibility to change the network, deploy the subgraph on launch, and run commands to populate it with data to test.

What do you think?

nivida: > Moving it to its own package (decoupling it from the server).

Do it :) You just have to move it to the root folder or to a new package folder 👍 

> Having the possibility to change the Ethereum RPC endpoint.

Could be done with arguments for the shell script and defining a default env value in the shell script 

> Having the possibility to run it without the server

The govern-server isn't included in the docker-compose :-) 

> deploy the subgraph on launch

Could be achieved by simply adding an additional command from existing ones to the root ``package.json``

> to change the network

Can also easily be done with arguments for the shell script (I think you mean the fork option right?)

> run commands to populate it with data to test

Yup, this is definitely something which should have been done in a separate PR cause it takes a bit more time :-)

kanaagawa: Maybe we can keep this section as it is as an advance setup. And first mention about the docker-compose file Sam create.

0xGabi (approved):
LGTM!

packages/govern-subgraph/README.md
bpierre: I was trying to edit the document to mention the two possibilities, but it was making it a bit complicated: it would require to tell people to go to adapt the server’s docker-compose (by deleting ganache and changing the Ethereum node), and the section would still need to explain how to deploy the subgraph.

I opened an issue to do as @nivida suggested next week https://github.com/aragon/govern/issues/122

PR #111, opened 2020-10-29
Core: GovernQueue test

### LINKED ISSUE #15: Core(GovernQueue): add tests
PR: https://github.com/aragon/govern/pull/111

nivida:
closes #15


nivida:
Note: I think we should/could move the ``GovernQueueStateLib`` into a separate file. This because the ``GovernQueue``file would look cleaner.

nivida:
Note: The complexity of the ``GovernQueue`` test is from my point of view a clear sign for bad OOD. I think we should think again about the separation of concerns closer. This would increase the possibility to re-use code, easn the audit process, possibly lower the deployment cost of an ERC3k DAO in many cases, and improve the situation if single pieces of a DAO should get improved/changed.

izqui:
I don't disagree, but also I wasn't optimizing for great OOD, nor I think we should.

Having a concise smart contract that can be read in one go without having to jump between multiple contracts is more important for me than easier testability. If a dev needs to change something, they only need to touch in one file without thinking about dependencies.

nivida:
> I don't disagree, but also I wasn't optimizing for great OOD, nor I think we should.
> 
> Having a concise smart contract that can be read in one go without having to jump between multiple contracts is more important for me than easier testability. If a dev needs to change something, they only need to touch in one file without thinking about dependencies.

Yes, I wouldn't do that just for better tests. But I thought there are probably also some other advantages.. 🤷‍♂️ 
@facuspagnuolo thought he will think about my feedback during his review :-)

nivida:
Note: Tests can't have all the assertions they should have because of this bug https://github.com/EthWorks/Waffle/issues/245

packages/govern-core/contracts/test/TestToken.sol
```diff
@@ -0,0 +1,179 @@
+/*
```
nivida, line 1: Use token contract from ``govern-token`` and add this package as ``devDependency``

packages/govern-core/test/pipelines/helpers.ts
```diff
@@ -0,0 +1,86 @@
+import { keccak256, defaultAbiCoder, solidityPack } from 'ethers/lib/utils'
```
nivida, line 1: Move those test-helpers to the ``govern-contract-utils`` package instead of duplicating it. (Don't forget to update the test in the ``erc3k`` package)

nivida: ... and add a unit-test test for those test-helpers cause of the complexity and importance they have.

nivida:
Note: https://github.com/aragon/govern/blob/master/packages/govern-core/contracts/pipelines/GovernQueue.sol#L112

Does not collect the tokens as expected. It should instead of collecting and sending the token to the ``msg.sender`` send it to the ``Queue`` contract. This means the deposit ``collectFrom`` and ``releaseTo`` methods of the ``DepositLib``need to be changed accordingly (e.g.: ``safeTransferFrom`` should be used in ``releaseTo``). Another possible solution would be to transfer the tokens to the ``ERC20`` but I don't think this is something we really want to do :-)

nivida:
Note: https://github.com/aragon/govern/blob/master/packages/erc3k/contracts/ERC3000Data.sol#L53

``this`` should probably for reasons of syntactical correctness be explicitly transformed to an address with ``address(this)``

izqui:
We should finish this ASAP @nivida

nivida (commented):
LGTM! Thanks for the finish @novaknole

PR #125, opened 2020-10-30
Server: use complete subgraph URLs in the networks

bpierre:


packages/govern-server/src/core/config.ts
```diff
@@ -1,22 +1,22 @@
...
+      'https://api.thegraph.com/subgraphs/name/aragon/aragon-govern-mainnet',
+  },
+  {
+    chainId: 4,
+    name: 'rinkeby',
+    ensAddress: '0x98df287b6c145399aaa709692c8d308357bc085d',
+    subgraphUrl:
+      'https://api.thegraph.com/subgraphs/name/aragon/aragon-govern-rinkeby',
+  },
+  // {
```
kanaagawa: Maybe we can keep a TODO to add xDai instead of keep the code commented?

0xGabi (approved):
LGTM! 🙌

packages/govern-server/src/core/config.ts
bpierre: Done https://github.com/aragon/govern/pull/125/commits/b9b5871d014273494f9a3407e65970ee9c04c1b0

PR #133, opened 2020-10-30
Fix "good first issue" link in README

onbjerg:
The link was dead :skull: because the label was renamed


bpierre:
Thanks Oliver :slightly_smiling_face:

PR #132, opened 2020-10-30
Create, core: Allow to create token from GovernBaseFactory

izqui:
**Note: Registered event has changed to add the token address cc @0xGabi @Evalir**

Please do the requires changes to the subgraph before merging. I will de-assign myself when done.

PR #137, opened 2020-10-30
Fix GovernToken tests

onbjerg:
The Solidity compiler was complaining that `memory` was used instead of `calldata` in `GovernTokenFactory#newToken`.

This made the tests fail on master - and this PR should fix that.﻿


nivida:
this got already fixed in #132 :-)

onbjerg:
I see!

PR #138, opened 2020-10-30
Server API: add missing resolvers

bpierre:


packages/govern-server/src/api-graphql/resolvers.ts
```diff
@@ -20,12 +20,22 @@ export default function resolvers(govern: GovernCore): IResolvers {
     Dao: {
       async registryEntries(parent) {
         return Promise.all(
-          parent.registryEntries.map(async ({ id }: { id: string }) =>
-            govern.registryEntry(id)
-          )
+          parent.registryEntries
+            .map(async ({ id }: { id: string }) => {
+              const entry = await govern.registryEntry(id)
```
nivida, line 25: This should be moved to the ``GovernCore`` and I still don't get the ``Dao.registryEntries`` cause it would be ``Registry.entries``. (already discussed it on discord quickly with @0xGabi)

packages/govern-server/src/api-graphql/resolvers.ts
```diff
@@ -20,12 +20,22 @@ export default function resolvers(govern: GovernCore): IResolvers {
...
+                ? {
+                    ...entry,
+                    executor: govern.dao(entry.executor.address),
+                  }
+                : null
+            })
+            .filter(Boolean)
         )
       },
+      async queues(parent) {
```
nivida, line 36: What exactly will this return? A DAO only has one queue but a queue with many container hashes stored.

packages/govern-server/src/core/thegraph-queries/registry.ts
```diff
@@ -13,20 +13,72 @@ const RegistryBase = gql`
...
+        }
+        challengeDeposit {
+          id
+          token
+          amount
+        }
+        resolver
+        rules
+      }
+      queued {
```
nivida, line 38: would rename ``queued`` to ``queue``  and isn't it possible to use ``fragments`` here? will lower the maintenance burden..

packages/govern-server/src/api-graphql/schema.ts
```diff
@@ -22,6 +22,7 @@ export default gql`
     registryEntries: [RegistryEntry!]!
     containers: [Container]!
     roles: [Role!]!
+    queues: [GovernQueue]!
```
nivida, line 25: A DAO only has one queue

nivida (changes requested):

packages/govern-server/src/api-graphql/resolvers.ts
bpierre: > This should be moved to the `GovernCore`

Moving it to `GovernCore` would mean querying all the registry entries and their data. The idea here is to find the right balance between querying everything in a giant query, or doing a big amount of small queries to resolve individual fields.

> I still don't get the `Dao.registryEntries` cause it would be `Registry.entries`. (already discussed on discord quickly with @0xGabi)

`RegistryEntry` contains a DAO (executor) and a queue, so we could have any amount of `RegistryEntry`. Let me know if that answers your question, I’m not sure it does :smile:

packages/govern-server/src/api-graphql/resolvers.ts
bpierre: This returns all the queues associated to the DAO via registry entries.

packages/govern-server/src/core/thegraph-queries/registry.ts
bpierre: That’s how it used to be, but we renamed it to avoid having `queue.queue`. We now only use `queue` to refer to a `GovernQueue`, and `queued` when it is a list of `Containers`.

packages/govern-server/src/api-graphql/schema.ts
bpierre: Do you mean that we should enforce this idea, only ever returning the first queue? Technically a DAO can have any number of queues.

packages/govern-server/src/api-graphql/resolvers.ts
nivida: I think ``parent.registryEntries`` can easily get passed to the ``GovernCore`` nothing to change at the request logic here.. just moving the stuff :-) 

> RegistryEntry contains a DAO (executor) and a queue, so we could have any amount of RegistryEntry. Let me know if that answers your question, I’m not sure it does

~~Yes, we could have any amount of registry entries in the whole system but we only have one entry per DAO. This as I see is a child of the Dao GraphQL Type which would be wrong :-)~~

... never mind forgot you are able to register different executor/queue combinations...

packages/govern-server/src/api-graphql/resolvers.ts
nivida: I see.. totally forgot you are able to register multiple DAO workflows 👌

packages/govern-server/src/api-graphql/schema.ts
nivida: ... sorry forgot that it is possible to have many different queues for an executor...

PR #153, opened 2020-10-30
Config: govern.js 'prepublishOnly'

nivida:
- adds ``prepublishOnly``
- ``files`` property in ``package.json`` defined


bpierre (approved):
All good, let’s see why it doesn’t build for me before merging.

PR #180, opened 2020-11-04
Utils(ACL): Implement ACL Oracle

### LINKED ISSUE #162: Utils(ACL): implement oracle checks

izqui:
Close #162


nivida:
Just stumbled across the [post](https://forum.aragon.org/t/about-acl-oracles/1805) of @facuspagnuolo and would like to know what the decision and thoughts were at the end related to ACL Oracles. I think this is something I have to know to be able to review this PR. Probably especially because I like the brought up points of Facu.

But I see that the example of the [post](https://forum.aragon.org/t/about-acl-oracles/1805) is I think not enough dynamic for the use case we have here but on the other hand, do I strongly like the point of Facu to not add more complexity to the ACL.

PR #181, opened 2020-11-04
Utils(SafeERC20): Ensure token address is a contract

izqui:
Otherwise if a provided token address﻿is an EOA, all transfers could be assumed to succeed

### DIFF
--- a/packages/govern-contract-utils/contracts/address-utils/AddressUtils.sol
+++ b/packages/govern-contract-utils/contracts/address-utils/AddressUtils.sol
@@ -15,7 +15,7 @@ library AddressUtils {
 
     function isContract(address addr) internal view returns (bool result) {
         assembly {
-            result := extcodesize(addr)
+            result := not(iszero(extcodesize(addr)))
         }
     }
 }
--- a/packages/govern-contract-utils/contracts/address-utils/AddressUtils.sol
+++ b/packages/govern-contract-utils/contracts/address-utils/AddressUtils.sol
@@ -12,4 +12,10 @@ library AddressUtils {
     function toAddress(address addr) internal pure returns (address payable) {
         return address(bytes20(addr));
     }
+
+    function isContract(address addr) internal view returns (bool result) {
+        assembly {
+            result := extcodesize(addr)
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


nivida (approved):

nivida:
CD run fails cause of the e2e tests for the govern-server.. will improve the situation there asap.

packages/govern-contract-utils/contracts/address-utils/AddressUtils.sol
```diff
@@ -12,4 +12,10 @@ library AddressUtils {
     function toAddress(address addr) internal pure returns (address payable) {
         return address(bytes20(addr));
     }
+
+    function isContract(address addr) internal view returns (bool result) {
+        assembly {
+            result := extcodesize(addr)
```
sohkai: Just to be safe, we might as well explicitly cast this to a bool. It's a small gas overhead, but feels better than leaving around the original uint256 size in the stack.

packages/govern-contract-utils/contracts/erc20/SafeERC20.sol
```diff
@@ -84,7 +91,6 @@ library SafeERC20 {
                 default { }
             }
         }
-        return ret;
     }
 
     function staticInvoke(address _addr, bytes memory _calldata)
```
sohkai: Not related to this specific PR, so better doing this separately, but we should remove the remaining `staticInvoke()` utilities since the compiler does this automatically now.

sohkai (approved):
LGTM.

However, depending on how we use (and expect to use) `SafeERC20` here, this is extra overhead (~800 gas) that could be checked elsewhere. For example, on the protocol, we most likely won't add this overhead because the token selection is protected in one way or another.

packages/govern-contract-utils/contracts/erc20/SafeERC20.sol
izqui: Sounds good, thanks for the review

packages/govern-contract-utils/contracts/address-utils/AddressUtils.sol
nivida: Because it is anyways just in the stack of that call do I think is it totally okay to not convert it to a boolean and to save this gas.

PR #190, opened 2020-11-04
CI/CD: Server and subgraph rinkeby on master commits, mainnet on v* tag

### LINKED ISSUE #174: Server: continuously deploy rinkeby instance
We currently have `govern.backend.aragon.org` that is updated on every version tag.
I will add a rinkeby instance `govern-rinkeby.backend.aragon.org` which will deploy continuously on every commit.

promaty:
With these manifests rinkeby subgraph/server will deploy on every master commit and mainnet on v* tag.
Also I moved server `Dockerfile` to the top so we install it using `yarn.lock` file.

closes #174

### DIFF
--- a/.github/workflows/server-ci-cd.yml
+++ b/.github/workflows/server-ci-cd.yml
@@ -12,7 +12,6 @@ env:
 
 jobs:
   build:
-    if: github.ref == 'refs/heads/master'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/server-ci-cd.yml
+++ b/.github/workflows/server-ci-cd.yml
@@ -1,19 +1,23 @@
 name: Server CI/CD
 on:
   push:
+    branches:
+    - master
     tags:
     - v*
 env:
-  # This is a base repository and we use ${GITHUB_REF##*/} to set the version of the container
+  # This is a base repository and we use git refs to set the version of the container
+  # ${GITHUB_REF##*/} will be either a branch name or a tag depending on the event
   REPO: aragon/govern-server
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


.github/workflows/server-ci-cd.yml
```diff
@@ -1,19 +1,23 @@
...
     - v*
 env:
-  # This is a base repository and we use ${GITHUB_REF##*/} to set the version of the container
+  # This is a base repository and we use git refs to set the version of the container
+  # ${GITHUB_REF##*/} will be either a branch name or a tag depending on the event
   REPO: aragon/govern-server
 
 jobs:
   build:
+    if: github.ref == 'refs/heads/master'
```
kanaagawa: I wonder if this condition is meet even on a push tag.

.github/workflows/subgraph-ci-cd.yml
```diff
@@ -11,7 +13,28 @@ jobs:
...
+    - name: Install node
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12
+
+    - name: Cache yarn modules
+      uses: actions/cache@v2
+      with:
+        path: '**/node_modules'
+        key: yarn-${{ hashFiles('yarn.lock') }}
```
kanaagawa, line 26: Nice!

0xGabi (approved):

.github/workflows/server-ci-cd.yml
promaty: Yes it is. The idea is to deploy to rinkeby on every master commit and to master when you add `v*` tag I think this makes sense?

promaty: good point, removed this part to build on tag as well

PR #193, opened 2020-11-05
fixes DepositLib.collectFrom

### LINKED ISSUE #184: Utils&Core: Fix DepositLib and update GovernQueue
The [GovernQueue](https://github.com/aragon/govern/blob/master/packages/govern-core/contracts/pipelines/GovernQueue.sol#L112
) does not collect the tokens as expected. It should instead of collecting and sending the token to the ``msg.sender`` send it to the ``Queue`` contract. This means the deposit ``collectFrom`` and ``releaseTo`` methods of the ``DepositLib``need to be changed accordingly (e.g.: ``safeTransferFrom`` should be used in ``releaseTo``).

nivida:
closes #184


Evalir (approved):
This seems like a super small change so we can merge ASAP. :)

nivida:
Yup, I will merge it now :-)

PR #195, opened 2020-11-05
Core(Queue): Update to use Aragon Protocol

izqui:


packages/govern-core/contracts/pipelines/GovernQueue.sol
```diff
@@ -71,8 +71,8 @@ contract GovernQueue is IERC3000, AdaptativeERC165, IArbitrable, ACL {
 
     function initialize(address _aclRoot, ERC3000Data.Config memory _initialConfig) public initACL(_aclRoot) onlyInit("queue") {
         _setConfig(_initialConfig);
-        _registerStandard(ARBITRABLE_INTERFACE_ID);
-        _registerStandard(ERC3000_INTERFACE_ID);
+        _registerStandard(type(IArbitrable).interfaceId);
```
sohkai: How does this work? Does Solidity automatically generate this for us?

But also, there should literally be no interfaces resulting from this. `IArbitrable` doesn't specify any functions to implement.

packages/govern-core/contracts/pipelines/GovernQueue.sol
```diff
@@ -244,7 +252,7 @@ contract GovernQueue is IERC3000, AdaptativeERC165, IArbitrable, ACL {
         return _execute(_container.payload, containerHash);
     }
 
-    function settleRejection(ERC3000Data.Container memory _container) public {
+    function settleRejection(ERC3000Data.Container memory _container) internal returns (bytes32, bytes[] memory){
```
sohkai: Since this is an internal function, could we not return anything here and handle the empty returns in the caller?

I don't think we should be trying to make type of thing look "nice" in Solidity :).

packages/govern-contract-utils/contracts/protocol/IArbitrable.sol
```diff
@@ -2,14 +2,17 @@
  * SPDX-License-Identifier:    MIT
```
sohkai: Based on our conversation earlier, I'm not sure if these make sense belonging in the utils. I imagine these are not being used outside of `govern-core`?

packages/govern-contract-utils/contracts/protocol/IArbitrable.sol
```diff
@@ -2,14 +2,17 @@
...
 
 import "./IArbitrator.sol";
 
+/**
+* @dev The Arbitrable instances actually don't require to follow any specific interface.
+*      Note that this is actually optional, although it does allow the Protocol to at least have a way to identify a specific set of instances.
+*/
 abstract contract IArbitrable {
-    bytes4 internal constant ARBITRABLE_INTERFACE_ID = bytes4(0x88f3ee69);
     uint256 internal constant ALLOW_RULING = 4;
```
sohkai: Perhaps we should move this constant into `GovernQueue`? Arbitrables choose this value themselves.

packages/govern-core/contracts/pipelines/GovernQueue.sol
```diff
@@ -185,22 +185,31 @@ contract GovernQueue is IERC3000, AdaptativeERC165, IArbitrable, ACL {
...
+            ? GovernQueueStateLib.State.Approved
+            : GovernQueueStateLib.State.Rejected;
+        queue[containerHash].checkAndSetState(
+            GovernQueueStateLib.State.Challenged, // checks now that this container was indeed in Challenged state
+            newState
+        );
 
-        if (state == GovernQueueStateLib.State.Approved) {
+        emit Resolved(containerHash, msg.sender, arbitratorApproved);
+        emit Ruled(arbitrator, _disputeId, ruling);
```
sohkai, line 212: It seems a bit weird we have two different events for this, but not sure if there's anything we can do.

On the protocol side, when we considered the `Ruled` event, it was only to set a standard event so that it'd be easier to find Arbitrables later from the Protocol's side (which we assume will be blind to ERC3k's interfaces / internals).

sohkai (commented):
Couple of notes for now, I think the logic looks good but will review again with the rest of the contracts.

packages/govern-contract-utils/contracts/protocol/IArbitrable.sol
izqui: As we talked, I think it makes sense to keep them in core unless we use some utils in several packages

sohkai: Done in b766b64.

sohkai:
@izqui @nivida I've added a few commits to polish a few rough edges.

The only logical change was with a0ba123, which fixes the submitter's collateral being stuck in the case of a veto on a challenged action. Like rejection, I felt it most appropriate to return this collateral to the challenger.

PR #197, opened 2020-11-05
Create: refactor

### LINKED ISSUE #175: Create: refactor
This issue aim to use hardhat advance functionalities:

- [ ] Move to typescript

- [ ] Add helpers to abstract task creation

- [ ] Migrate to hardhat

0xGabi:
Close #175 

The changes in this PR include:
- Etherscan verification using a retry mechanisim
- Migrate package to typescript + hardhat
- Use `hardhat-deploy` as deployment tool

---
Note: Depends on https://github.com/ethereum-ts/TypeChain/pull/300 ✅


packages/govern-create/helpers/etherscan-verification.ts
```diff
@@ -0,0 +1,85 @@
+import { exit } from 'process'
+import fs from 'fs'
+import { file } from 'tmp-promise'
```
bpierre: Should we name this `tmpFile` for clarity?

```suggestion
import { file as tmpFile } from 'tmp-promise'
```

packages/govern-create/helpers/etherscan-verification.ts
```diff
@@ -0,0 +1,85 @@
...
+      `Current network ${currentNetwork} not supported. Please change to one of the next networks: ${SUPPORTED_ETHERSCAN_NETWORKS.toString()}`
+    )
+  }
+
+  try {
+    logWarn(
+      'Delaying Etherscan verification due their API can not find newly deployed contracts'
+    )
+    const msDelay = 3000
+    const times = 30
```
bpierre: Should we move these up?

packages/govern-create/helpers/etherscan-verification.ts
```diff
@@ -0,0 +1,85 @@
...
+      await HRE.run(task, params)
+      cleanup()
+    } else {
+      cleanup()
+      logError(
+        'Errors after all the retries, check the logs for more information.'
+      )
+    }
+  } catch (error) {
+    counter--
```
bpierre: I think we could pass `timer-1` to `runTaskWithRetry` rather than having `counter`?

packages/govern-create/helpers/etherscan-verification.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  try {
+    if (times) {
+      await HRE.run(task, params)
+      cleanup()
+    } else {
+      cleanup()
+      logError(
+        'Errors after all the retries, check the logs for more information.'
+      )
```
bpierre: I think this could move in the `catch`?

packages/erc3k/test/erc3k-data.test.ts
```diff
@@ -100,25 +103,29 @@ describe.skip('ERC3000Data', function () {
...
   })
 
   it('calls testContainerHash and returns the expected hash', async () => {
-    expect(
-      await erc3kDataLib.testContainerHash(container)
-    ).to.be.equal(
+    expect(await erc3kDataLib.testContainerHash(container)).to.be.equal(
       keccak256(
         solidityPack(
           ['string', 'address', 'uint', 'bytes32', 'bytes32'],
```
nivida: I think the first array should have breaks as well to keep the consistency.

nivida: ... and as a hint: Those changes are out of the scope of this PR and we probably should actually just do stuff if it is clearly in the scope of this PR. This will improve the situation for reviewers to actually see the changes described in the PR description.

packages/erc3k/test/erc3k-exec.test.ts
```diff
@@ -1,5 +1,8 @@
 import { ethers } from '@nomiclabs/buidler'
-import { Erc3000ExecutorMock, Erc3000ExecutorMockFactory } from '../typechain'
+import {
+  Erc3000ExecutorMock,
+  Erc3000ExecutorMock__factory as Erc3000ExecutorMockFactory,
```
nivida: Is this the way ``typechain`` fixed the naming issue? or why exactly are the names now like this?

packages/erc3k/test/erc3k-data.test.ts
```diff
@@ -100,25 +103,29 @@ describe.skip('ERC3000Data', function () {
   })
 
   it('calls testConfigHash and returns the expected hash', async () => {
-    expect(
-      await erc3kDataLib.testConfigHash(container.config)
-    ).to.be.equal(getConfigHash())
+    expect(await erc3kDataLib.testConfigHash(container.config)).to.be.equal(
```
nivida: Would keep this on one line or keep the code style as it was before. But this doesn't really look good from my point of view. :-)

packages/govern-contract-utils/test/acl.ts
```diff
@@ -1,21 +1,18 @@
...
-  Acl,
-  AclFactory
-} from '../typechain'
+import { Acl, Acl__factory as AclFactory } from '../typechain'
 
 const ERRORS = {
   AUTH: 'acl: auth',
   FROZEN: 'acl: frozen',
-  BAD_FREEZE: 'acl: bad freeze'
+  BAD_FREEZE: 'acl: bad freeze',
```
nivida: Sorry, my fault will add the prettier extension in my IDE to not always break the code style.

packages/govern-contract-utils/test/acl.ts
```diff
@@ -43,11 +40,18 @@ describe('ACL', function () {
...
 
   const grant = (inst: Acl, role = ROLE, who = notRoot) => inst.grant(role, who)
-  const revoke = (inst: Acl, role = ROLE, who = notRoot) => inst.revoke(role, who)
+  const revoke = (inst: Acl, role = ROLE, who = notRoot) =>
+    inst.revoke(role, who)
   const freeze = (inst: Acl, role = ROLE) => inst.freeze(role)
 
-  const assertRole = async (shouldHaveRole = true, role = ROLE, who = notRoot) =>
-    expect(await acl.roles(role, who)).to.equal(shouldHaveRole ? ALLOW_ROLE : NO_ROLE)
+  const assertRole = async (
```
nivida: let us just use the ``function`` keyword here if we anyways format it like this.

packages/govern-contract-utils/test/acl.ts
```diff
@@ -129,17 +133,19 @@ describe('ACL', function () {
...
-  
+
     it('non-root cannot freeze', async () => {
       await expect(freeze(aclNotRoot)).to.be.revertedWith(ERRORS.AUTH)
     })
   })
 
   it('root cannot freeze by granting', async () => {
-    await expect(grant(acl, ROLE, FREEZE_ADDR)).to.be.revertedWith(ERRORS.BAD_FREEZE)
+    await expect(grant(acl, ROLE, FREEZE_ADDR)).to.be.revertedWith(
```
nivida: I think it really looks inconsistent to have only one test case with this break. Is this a prettier config?

packages/govern-create/deployed-contracts.json
```diff
@@ -0,0 +1,58 @@
...
+  },
+  "GovernBaseFactory": {
+    "hardhat": {
+      "address": "0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e"
+    },
+    "rinkeby": {
+      "address": "0xD6bC44402b42B47562E8E4E9822722Dd9ec86ac3"
+    }
+  }
+}
```
nivida: EOF

packages/govern-create/hardhat.config.ts
```diff
@@ -0,0 +1,59 @@
...
+import 'hardhat-typechain'
+import dotenv from 'dotenv'
+import path from 'path'
+import fs from 'fs'
+import { logInfo } from './helpers/logger'
+
+dotenv.config({ path: '../../.env' })
+
+if (!process.env.SKIP_TASKS) {
+  ;['misc', 'deployments', 'sequences'].forEach((folder) => {
```
nivida: Don't think this semicolon is required.

packages/govern-create/hardhat.config.ts
```diff
@@ -0,0 +1,59 @@
...
+    const tasksPath = path.join(__dirname, 'tasks', folder)
+    fs.readdirSync(tasksPath).forEach((task) => require(`${tasksPath}/${task}`))
+  })
+  logInfo('Loaded tasks')
+}
+
+const HARDFORK = 'istanbul'
```
nivida: I think we can move this to the env file as well.

packages/govern-create/hardhat.config.ts
```diff
@@ -0,0 +1,59 @@
...
+const ETHERSCAN_KEY = process.env.ETHERSCAN_KEY || ''
```
nivida: Those ``const`` aren't really required. I think we can directly use the values and conditional statements in the config object :-)

packages/govern-create/helpers/artifactsDb.ts
```diff
@@ -0,0 +1,20 @@
...
+const adapter = new FileSync('./deployed-contracts.json')
+export const getDb = () => low(adapter)
+
+export const registerContractInJsonDb = async (
+  contractId: eContractid,
+  contractAddress: Address
+) => {
+  const currentNetwork = HRE.network.name
+
+  await getDb()
```
nivida: I think the breaks here are wrong or the breaks are wrong in the above already commented files.

packages/govern-create/helpers/delploys.ts
```diff
@@ -0,0 +1,131 @@
+import {
```
nivida: I think the file name is wrong.

packages/govern-create/helpers/delploys.ts
```diff
@@ -0,0 +1,131 @@
...
+  return instance
+}
+
+export const deployGoverQueueFactory = async (
+  network: string,
+  verify?: boolean
+) => {
+  const id = eContractid.GovernQueueFactory
+
+  const instance = await deployContract<GovernQueueFactory>(id, [])
```
nivida: I think this could also be written as: 
```
const instance = (await (await deployContract<GovernQueueFactory>(id, [])).deployTransaction.wait())
```

packages/govern-create/helpers/delploys.ts
```diff
@@ -0,0 +1,131 @@
...
+    logDeploy(id, network, instance.address)
+  }
+
+  return instance
+}
+
+export const deployGoverQueueFactory = async (
+  network: string,
+  verify?: boolean
+) => {
```
nivida: return type is missing

packages/govern-create/helpers/delploys.ts
```diff
@@ -0,0 +1,131 @@
...
+} from '../typechain'
+import { verifyContract } from './etherscan-verification'
+import { deployContract, NO_TOKEN, TWO_ADDRR, ZERO_ADDR } from './helpers'
+import { logDeploy } from './logger'
+import { eContractid } from './types'
+
+export const deployGoverRegistry = async (
+  network: string,
+  verify?: boolean
+) => {
```
nivida: return type is missing

packages/govern-create/helpers/artifactsDb.ts
```diff
@@ -0,0 +1,20 @@
...
+import { HRE } from './helpers'
+import { Address, eContractid } from './types'
+
+const adapter = new FileSync('./deployed-contracts.json')
+export const getDb = () => low(adapter)
+
+export const registerContractInJsonDb = async (
+  contractId: eContractid,
+  contractAddress: Address
+) => {
```
nivida: return type is missing (``void``)

packages/govern-create/helpers/delploys.ts
nivida: I also see a lot of duplicated code here which wouldn't exist if this would be done with a ``Deployer`` class and if you want to escalate completely from a clean code perspective could you also create a ``AbstractDeployer`` etc. ;-P

packages/govern-create/helpers/delploys.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  if (network !== 'hardhat' && network !== 'soliditycoverage') {
+    logDeploy(id, network, instance.address)
+    logDeploy(eContractid.Queue, network, queueBase)
+  }
+
+  return instance
+}
+
+export const deployGoverFactory = async (network: string, verify?: boolean) => {
```
nivida: return type missing

packages/govern-create/helpers/delploys.ts
nivida: This could be done for all functions. But one argument against it is the readability of it. So feel free to choose.

packages/govern-create/helpers/delploys.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  if (network !== 'hardhat' && network !== 'soliditycoverage') {
+    logDeploy(id, network, instance.address)
+  }
+
+  return instance
+}
+
+export const deployGoverBaseFactory = async (
+  [registry, governFactory, queueFactory, tokenFactory]: [
```
nivida: I would create a small type definition here in this file instead of typing this more complex type in the function arguments. A simple value type definition as ``string``, ``number`` etc. are good. Moving this type definition up with an explanatory name will definitely increase the readibility.

packages/govern-create/helpers/etherscan-verification.ts
```diff
@@ -0,0 +1,85 @@
+import { exit } from 'process'
+import fs from 'fs'
+import { file } from 'tmp-promise'
+import { HRE } from './helpers'
+import { logError, logInfo, logWarn } from './logger'
+
+export const SUPPORTED_ETHERSCAN_NETWORKS = ['main', 'rinkeby']
+
+function delay(ms: number) {
```
nivida: Return type is missing

nivida: .. on all other places as well

packages/govern-create/helpers/etherscan-verification.ts
```diff
@@ -0,0 +1,85 @@
+import { exit } from 'process'
+import fs from 'fs'
+import { file } from 'tmp-promise'
+import { HRE } from './helpers'
+import { logError, logInfo, logWarn } from './logger'
+
+export const SUPPORTED_ETHERSCAN_NETWORKS = ['main', 'rinkeby']
```
nivida: This is a configuration and should be in the same place as all other configurations.

packages/govern-create/helpers/etherscan-verification.ts
```diff
@@ -0,0 +1,85 @@
...
+
+function delay(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms))
+}
+
+export const verifyContract = async (
+  address: string,
+  constructorArguments: any[]
+) => {
+  const currentNetwork = HRE.network.name
```
nivida: I would recommend from a software architectural standpoint to pass such values in the right place instead of defining them here in the function.

packages/govern-create/helpers/etherscan-verification.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  if (!process.env.ETHERSCAN_KEY) {
+    throw Error('Missing process.env.ETHERSCAN_KEY.')
+  }
+  if (!SUPPORTED_ETHERSCAN_NETWORKS.includes(currentNetwork)) {
```
nivida: Such configuration validations could be done within a ``Configuration``VO object and would prevent code duplication.

packages/govern-create/helpers/etherscan-verification.ts
```diff
@@ -0,0 +1,85 @@
...
+  return new Promise((resolve) => setTimeout(resolve, ms))
+}
+
+export const verifyContract = async (
+  address: string,
+  constructorArguments: any[]
+) => {
+  const currentNetwork = HRE.network.name
+
+  if (!process.env.ETHERSCAN_KEY) {
```
nivida: Same here.. just pass the value and do the validation of such env configuration on the place this function is called.

packages/govern-create/helpers/delploys.ts
```diff
@@ -0,0 +1,131 @@
...
+    string,
+    string,
+    string,
+    string
+  ],
+  network: string,
+  verify?: boolean
+) => {
+  const id = eContractid.GovernBaseFactory
+  const args = [registry, governFactory, queueFactory, tokenFactory]
```
nivida: Don't think this is required to do.

packages/govern-create/helpers/etherscan-verification.ts
```diff
@@ -0,0 +1,85 @@
...
+    const msDelay = 3000
+    const times = 30
+    // Write a temporal file to host complex parameters for hardhat-etherscan https://github.com/nomiclabs/hardhat/tree/master/packages/hardhat-etherscan#complex-arguments
+    const { fd, path, cleanup } = await file({
+      prefix: 'verify-params-',
+      postfix: '.js',
+    })
+    fs.writeSync(
+      fd,
+      `module.exports = ${JSON.stringify([...constructorArguments])};`
```
nivida: Why aren't we have this in a config? do those arguments change with each deployment from our side?

packages/govern-create/helpers/etherscan-verification.ts
```diff
@@ -0,0 +1,85 @@
...
+    )
+
+    const params = {
+      address: address,
+      constructorArgs: path,
+    }
+    await runTaskWithRetry('verify', params, times, msDelay, cleanup)
+  } catch (error) {}
```
nivida: Are you sure you want to ignore any possible error thrown in this function?

packages/govern-create/helpers/etherscan-verification.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+  } catch (error) {
+    counter--
+    logInfo(`Retrying attemps: ${counter}.`)
+    logError(error.message)
+    await runTaskWithRetry(task, params, counter, msDelay, cleanup)
+  }
+}
+
+export const checkVerification = () => {
```
nivida: We have this check already above in ``verifyContract`` which does duplicate this and I'm anyways don't think we want to do that like this :-)

Have you thought to create a verifier object with self-explanatory private methods to increase readability?

packages/govern-create/helpers/helpers.ts
```diff
@@ -0,0 +1,127 @@
...
+export const writeObjectToFile = async (path: string, obj: object) =>
+  await promises.writeFile(path, JSON.stringify(obj))
+
+// Hardhat Runtime Environment
+export let HRE: HardhatRuntimeEnvironment = {} as HardhatRuntimeEnvironment
+export const setHRE = (_HRE: HardhatRuntimeEnvironment) => {
+  HRE = _HRE
+}
+
+export const ZERO_ADDR = `0x${'00'.repeat(20)}`
```
nivida: Those ``const`` aren't really helpers.. those are config properties aka default values.

packages/govern-create/helpers/helpers.ts
```diff
@@ -0,0 +1,127 @@
...
+  GovernRegistry,
+  GovernTokenFactory,
+} from '../typechain'
+
+export const writeObjectToFile = async (path: string, obj: object) =>
+  await promises.writeFile(path, JSON.stringify(obj))
+
+// Hardhat Runtime Environment
+export let HRE: HardhatRuntimeEnvironment = {} as HardhatRuntimeEnvironment
+export const setHRE = (_HRE: HardhatRuntimeEnvironment) => {
```
nivida: Why is this actually required?

packages/govern-create/helpers/helpers.ts
```diff
@@ -0,0 +1,127 @@
...
+import { getDb } from './artifactsDb'
+import {
+  GovernBaseFactory,
+  GovernFactory,
+  GovernQueueFactory,
+  GovernRegistry,
+  GovernTokenFactory,
+} from '../typechain'
+
+export const writeObjectToFile = async (path: string, obj: object) =>
```
nivida: Is it possible to have the same shorthand function syntax as below?

packages/govern-create/helpers/helpers.ts
```diff
@@ -0,0 +1,127 @@
+import { HardhatRuntimeEnvironment } from 'hardhat/types'
```
nivida: Return types are missing

packages/govern-create/helpers/logger.ts
```diff
@@ -0,0 +1,46 @@
...
+  color?: typeof ForegroundColor
+): string {
+  if (color) tag = chalk[color](tag)
+  return lines
+    .split('\n')
+    .map((line) => tag + line)
+    .join('\n')
+}
+
+export function logInfo(data: string): void {
```
nivida: I would create a simple static class for such a logger and use log levels you can pass instead of exposing different functions. This will definitely lower the code duplication here.

packages/govern-create/helpers/logger.ts
```diff
@@ -0,0 +1,46 @@
+import chalk, { ForegroundColor } from 'chalk'
```
nivida: I'm against using an additional dependency just to define a color in such an important place. Just pass the colors as documented. Any additional dependency means an additional security risk.

packages/govern-create/helpers/types.ts
```diff
@@ -0,0 +1,19 @@
+export type Address = string
+
+export enum eEthereumNetwork {
+  rinkeby = 'rinkeby',
+  main = 'main',
+  hardhat = 'hardhat',
+}
+
+export enum eContractid {
```
nivida: I would recommend to define it like this ``const enum eContractId{...}`` the generated JS is much smaller.

I would also not name those like this.. just call it ``CONTRACT_ID`` or ``CONTRACT_NAMES``. ``eEthereumNetwork`` should for example also be called ``ETHEREUM_NETWORK``.

nivida: I also don't think ``types.ts`` is the right place to define enum's cause they aren't just types :-)

packages/govern-create/tasks/sequences/deploy-factory.ts
```diff
@@ -0,0 +1,57 @@
...
+          tokenFactory.address,
+        ],
+        currentNetwork,
+        verify
+      )
+
+      const queueBase = await queueFactory.base()
+      const governBase = await governFactory.base()
+
+      await registerContractInJsonDb(GovernQueueFactory, queueFactory.address)
```
nivida: A clear sign to have a loop somewhere.

packages/govern-create/tasks/sequences/deploy-factory.ts
```diff
@@ -0,0 +1,57 @@
...
+const {
+  GovernBase,
+  GovernBaseFactory,
+  GovernFactory,
+  GovernTokenFactory,
+  QueueBase,
+  GovernQueueFactory,
+} = eContractid
+
+task('deploy-factory', 'Deploys an GovernBaseFactory instance')
```
nivida: Why not cover those tasks with simple and well-named ``Deployer`` object which will increase the readability and lower the code duplications.

packages/govern-create/tasks/sequences/deploy-govern.ts
```diff
@@ -0,0 +1,91 @@
...
+      if (verify && !useProxies) {
+        const dummyConfig = {
+          executionDelay: '0',
+          scheduleDeposit: [NO_TOKEN, '0'],
+          challengeDeposit: [NO_TOKEN, '0'],
+          resolver: ZERO_ADDR,
+          rules: '0x',
+        }
+
+        await verifyContract(queueAddress, [baseFactory.address, dummyConfig])
```
nivida: ``verifyContracts(any[])``

packages/govern-create/tasks/sequences/deploy-govern.ts
```diff
@@ -0,0 +1,91 @@
...
+      const args = logs
+        .filter(({ address }) => address === registry.address)
+        .map((log) => registry.interface.parseLog(log))
+        .find(({ name }) => name === 'Registered')
+
+      const queueAddress = args?.args[1] as string
+      const governAddress = args?.args[0] as string
+
+      if (verify && !useProxies) {
+        const dummyConfig = {
```
nivida: This config is duplicated.

packages/govern-create/tasks/sequences/deploy-govern.ts
```diff
@@ -0,0 +1,91 @@
...
+      )
+
+      const { logs } = await tx.wait()
+
+      const args = logs
+        .filter(({ address }) => address === registry.address)
+        .map((log) => registry.interface.parseLog(log))
+        .find(({ name }) => name === 'Registered')
+
+      const queueAddress = args?.args[1] as string
```
nivida: Why not define a type at ``const args``?

packages/govern-create/tasks/sequences/deploy-govern.ts
```diff
@@ -0,0 +1,91 @@
...
+        : await getGovernBaseFactory()
+
+      const tx = await baseFactory.newGovernWithoutConfig(
+        name,
+        token,
+        tokenName || name,
+        tokenSymbol,
+        useProxies,
+        {
+          gasLimit: useProxies ? 2e6 : 9e6,
```
nivida: A cool fancy way to write numbers but would recommend to write them out and probably have those in a configuration or at least as process env.

packages/govern-create/tasks/sequences/deploy-govern.ts
```diff
@@ -0,0 +1,91 @@
...
+        useProxies = true,
+        name,
+        verify,
+        token = NO_TOKEN,
+        tokenName = name,
+        tokenSymbol = 'GOV',
+      },
+      HRE
+    ) => {
+      setHRE(HRE)
```
nivida: Those to lines do get duplicated in each task.

packages/govern-create/tasks/sequences/deploy-registry.ts
```diff
@@ -0,0 +1,18 @@
...
+
+task('deploy-registry', 'Deploys an GoverRegistry instance')
+  .addFlag('verify', 'Verify the contracts via Etherscan API')
+  .setAction(async ({ verify }, HRE) => {
+    setHRE(HRE)
+    const currentNetwork = HRE.network.name
+
+    const governRegistry = await deployGoverRegistry(currentNetwork, verify)
+
+    await registerContractInJsonDb(GovernRegistry, governRegistry.address)
```
nivida: Where and for what exactly is this DB required?

packages/govern-create/test/base.ts
```diff
@@ -0,0 +1,63 @@
+import HRE, { waffle } from 'hardhat'
```
nivida: I will review this test after.

nivida (changes requested):
I found a lot of typescript related stuff and has many architectural improvements we can talk about. The main problem I see is that there is no architectural pattern to see which leads to many duplicated code and uncertainties. 

I will after the discussed changes are applied to do a functional test. I for now just looked at code style, architecture, and language-specific improvements.

packages/govern-contract-utils/test/acl.ts
bpierre: Yes probably, the limit is 80 chars.

packages/govern-create/hardhat.config.ts
bpierre: I think prettier always adds a semicolon when a statement starts with `[]`.

nivida: True that.. cause we do not have semicolons at all.. I normally do use semicolons can't see the reason to not use them cause they got "added" anyways later :)

packages/govern-create/helpers/types.ts
bpierre: Agreed, I only learned recently about the difference between `enum` and `const enum` :smile:  

Also I don’t think we need string values?

nivida: Agreed! :)

0xGabi:
@bpierre @nivida thanks for the comments I keep digging into hardhat and I finally decide to go with the `hardhat-deploy` plugin and it made things much simpler. I decide to open a new PR. This was replace by #255

PR #199, opened 2020-11-05
Remove erc3kjs

bpierre:
Removes the erc3kjs package.


0xGabi (approved):
LGTM!

Evalir (approved):
LGTM—But I have a point of discussion. To keep it short: should we remove this package? Sam has been working on `govern.js`, and so this package doesn't make much sense right now. Unless @izqui has some thoughts on where could we take this, I believe we shouldn't commit to maintain this.

bpierre:
@Evalir Agreed, I was thinking of keeping it in a minimal state for now, but we might as well remove it. We’ll still have a erc3k.js library later, once we’ll extract the ERC3k features from Govern.js.

nivida (approved):

PR #200, opened 2020-11-05
Subgraph: fix paths + add a build script

bpierre:
The build script runs `graph codegen` and `graph build` at once. It also displays an error message if the manifest files and `subgraph.yaml` don’t exist, explaining how to generate them.

The package paths in the templates are now resolved using the Node algorithm.

### DIFF
--- a/packages/govern-subgraph/scripts/build-manifest.sh
+++ b/packages/govern-subgraph/scripts/build-manifest.sh
@@ -11,6 +11,8 @@ fi
 
 DATA=manifest/data/$FILE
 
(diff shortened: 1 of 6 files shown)


Evalir (approved):
🔥🔥 Let's merge ASAP!

packages/govern-subgraph/scripts/build-manifest.sh
```diff
@@ -11,11 +11,15 @@ fi
 
 DATA=manifest/data/$FILE
 
+GOVERN_CORE_MODULE=$(node -e 'console.log(require("path").dirname(require.resolve("@aragon/govern-core/package.json")))')
```
kanaagawa, line 14: I didn't know you can do that 🧙‍♂️🔮

PR #203, opened 2020-11-06
Govern test

### LINKED ISSUE #14: Test Govern.sol

nivida:
closes #14


packages/govern-core/test/govern.test.ts
```diff
@@ -39,6 +44,45 @@ describe('Govern', function () {
...
+
+
+    await expect(signers[0].sendTransaction(
+      {
+        to: govern.address,
+        data: callbackSig
+      }
+      )
+    )
+    .to.emit(govern, EVENTS.RECEIVED_CALLBACK)
```
kanaagawa, line 82: Not sure if possible with a low level return, but maybe we can also assert the right `magicNumber` got returned.

0xGabi (approved):
LGTM 🙌

packages/govern-core/test/govern.test.ts
nivida: I think we can add this test case in the ``AdaptativeERC165`` test where it actually belongs to or we would duplicate the test cases too much. I was just not able to mock the extended contract.

PR #204, opened 2020-11-06
Core: GovernQueue.veto enhancement

### LINKED ISSUE #12: Allow veto on challenged items

nivida:
closes #12 

(This PR has ``test/govern-queue`` as base branch because it is required over there to finish the test without writing cases twice; this branch also has to wait on the PR #195 cause of the new arbitrator interface)


nivida:
To implement this accordingly do we have to change the ``disputeItemCache`` mapping cause we do not have a way to get the used arbitrator. This would also simplify the ``veto`` contract method signature and implementation cause overloading wouldn't be required. The required mapping would be ``containerhash->arbitrator``.

Edit:
Discussed with Jorge.. PR #195 will fix this 👌

PR #205, opened 2020-11-06
Add GovernLib

bpierre:
This is the library that used to be in `src/core/` in the server.


packages/govern-lib/src/utils/NetworkContext.ts
```diff
@@ -0,0 +1,96 @@
+import { Address, Networkish, Network } from '../types'
+import { networkContexts } from '../config'
+import { ErrorInvalidNetwork } from '../errors'
+
+class NetworkContext implements Network {
```
bpierre, line 5: @0xGabi I refactored a little bit the `Network` class to separate the idea of a “network” (as in Ethereum network) and a “network context” (Ethereum network + contextual data, i.e. the subgraph URL). Let me know what you think!

packages/govern-lib/README.md
```diff
@@ -0,0 +1,27 @@
+# GovernLib
+
+GovernLib is a library that allows Node.js apps and browsers to access the Govern related data in a unified way.
```
nivida, line 3: This sentence could confuse devs cause they aren't sure anymore if they should use the lib or govern.js.. We probably could call this package also ``govern-shared`` 🤷‍♂️

packages/govern-lib/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@aragon/govern-lib",
+  "version": "1.0.0",
+  "main": "dist/index.js",
+  "typings": "dist/index.d.ts",
+  "module": "dist/govern-lib.esm.js",
+  "license": "LGPL-3.0-or-later",
+  "files": ["dist"],
+  "scripts": {
+    "start": "tsdx watch",
```
nivida, line 12: Are we sure to use here tsdx and anywhere else just tsc? I think we should have a consistent build pipeline for all JS related packages.

packages/govern-lib/src/GovernLib.ts
```diff
@@ -0,0 +1,136 @@
...
+  QUERY_QUEUES_BY_DAO,
+} from './thegraph-queries'
+
+export type GovernLibConfig = {
+  network: Networkish
+  subgraphUrl?: string
+  verbose?: boolean
+}
+
+class GovernLib {
```
nivida, line 27: I would just call it Govern

packages/govern-lib/src/GovernLib.ts
```diff
@@ -0,0 +1,136 @@
...
+  #gql: GraphqlClient
+  readonly config: GovernLibConfig
+  readonly networkContext: NetworkContext
+
+  constructor(config: GovernLibConfig) {
```
nivida, line 32: Would be great to have funcDocs for all methods. It increases the readability especially if always a short helpful comment is existing.

packages/govern-lib/src/GovernLib.ts
```diff
@@ -0,0 +1,136 @@
...
+    this.#gql = new GraphqlClient(this.networkContext.subgraphUrl, {
+      verbose: config.verbose,
+    })
+  }
+
+  private async fetchResult<R>(
+    queryAndParams:
+      | [any]
+      | [
+          any, // TODO: use the type returned by gql``
```
nivida: ``DocumentNode``

packages/govern-lib/src/config.ts
```diff
@@ -0,0 +1,27 @@
+import { ChainId, Network } from './types'
+
+export const networkContexts = new Map<
```
nivida, line 3: I would just call this ``config`` or ``networks`` and rename the file.

packages/govern-lib/src/config.ts
```diff
@@ -0,0 +1,27 @@
...
+  ChainId,
+  Network & { subgraphUrl: string }
+>([
+  [
+    1,
+    {
+      chainId: 1,
+      name: 'ethereum',
+      ensAddress: '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e',
+      subgraphUrl:
```
nivida, line 13: Don't think this break is required.

packages/govern-lib/src/utils/NetworkContext.ts
```diff
@@ -0,0 +1,96 @@
...
+  readonly chainId: number
+  readonly ensAddress: Address
+  readonly name: string
+  readonly subgraphUrl: string
+
+  static chainIdFromName(name: string): number | undefined {
```
nivida, line 11: I think the ``constructor`` should always be on top of the class also if static factory methods do exist.

packages/govern-lib/src/utils/NetworkContext.ts
```diff
@@ -0,0 +1,96 @@
...
+    chainId,
+    ensAddress,
+    name,
+    subgraphUrl,
+  }: {
+    chainId?: number
+    ensAddress?: Address
+    name?: string
+    subgraphUrl?: string
+  }) {
```
nivida, line 46: Totally not a fan of such constructors. A constructor itself should not have such logic. Please create methods for this.

nivida (changes requested):

packages/govern-lib/README.md
bpierre: True, I will rephrase it. `govern-shared` doesn’t sound like something you could run from an app directly. I was trying to find a name similar to `govern-core`, that we can’t use since we already do for the contracts.

packages/govern-lib/package.json
bpierre: We started using tsdx in our other projects (use-wallet, use-token and use-viewport are using it now) since [the features it provides by default](https://github.com/formium/tsdx#features) are really close to what we want. I think we should keep moving every new TS library to it, or do you see a reason why not?

packages/govern-lib/src/config.ts
bpierre: Prettier :-)

packages/govern-lib/src/utils/NetworkContext.ts
bpierre: I was trying to separate static methods from instance ones, but I don’t have any strong opinion, let’s do this :+1:

packages/govern-lib/src/utils/NetworkContext.ts
bpierre: No strong opinion, but I find it easier to read when property assignments are done in the constructor, without having to jump to other methods. Is it what you are disagreeing with?

packages/govern-lib/src/GovernLib.ts
bpierre: The idea with GovernLib is to avoid using the term Govern, so we can refer to this particular project using a specific name rather than a too generic one.

packages/govern-lib/src/config.ts
bpierre: True let’s do this. I was trying to make the separation between network and network + subgraph (i.e. network context) more clear, but it is probably unnecessary for this file yes :+1:

packages/govern-lib/src/GovernLib.ts
bpierre: True, will add

packages/govern-lib/src/GovernLib.ts
bpierre: I remember having an issue with `DocumentNode` here at the time this TODO was written, will try again :+1:

packages/govern-lib/package.json
nivida: I couldn’t see the requirement for it but if a transition is ongoing then definitely :)

packages/govern-lib/src/utils/NetworkContext.ts
nivida: I have probably sometimes a to strong opinion in such cases 🙈😜

packages/govern-lib/src/utils/NetworkContext.ts
nivida: Yeah, I just learned from OOP that no logic should be in the constructors. I think accessors are a great use case for something like this :-)

packages/govern-lib/README.md
nivida: I'm not sure if we really want to let our users directly use the ``core`` functionalities of the ``govern-server`` package. I think they should use them through the ``govern-server`` package with a thin abstraction layer. This would give us a lot of freedom in the future of how we want to have our ``core`` functionalities be architected.

packages/govern-lib/src/GovernLib.ts
nivida: Probably we just want to have a ``GraphQLClient`` class in the core with the ``fetchResult`` method and we use this/extend from it in the ``govern-server`` package. We could for example also change the interface of this class in the ``core`` to be aligned with the ``ClientInterface`` the ``govern.js`` package does have. The big advantage we would have with it is it would lower the code duplication in the JS side of the ``Govern`` project. This means we would have to bring together the ``fetchResult`` and ``performQuery`` method of the currently named ``GovernLib`` and ``GraphqlClient`` object of the server or we just use the client object from the ``govern.js`` package.

packages/govern-lib/src/thegraph-queries/index.ts
```diff
@@ -0,0 +1,3 @@
+export { QUERY_DAO, QUERY_DAOS } from './dao'
```
nivida: Btw.: Great to have all those types and queries defined here for the later more decentralized version of Govern! :-)

... I'm not sure if we need here for an internal lib a ``index`` file if we just can import the required queries with an explicit path.

packages/govern-lib/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "@size-limit/preset-small-lib": "^4.7.0",
+    "husky": "^4.3.0",
+    "size-limit": "^4.7.0",
+    "tsdx": "^0.14.1",
+    "tslib": "^2.0.3",
+    "typescript": "^4.0.5"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "tsdx lint"
```
kanaagawa, line 30: Nice!

packages/govern-lib/package.json
kanaagawa: tsdx looks nice, I like the decision of moving toward it!

packages/govern-lib/README.md
kanaagawa: I agree with Sam that at this point maybe allowing user to use `core` maybe a bit overkill and could potentially limit our freedom to move fast and iterate quick on the `core` functionalities.

packages/govern-lib/tsconfig.json
```diff
@@ -0,0 +1,35 @@
+{
```
kanaagawa, line 1: Thanks, the comments were helpful to better understand some of the rules.

packages/govern-lib/tsconfig.json
```diff
@@ -0,0 +1,35 @@
...
+    "rootDir": "./src",
+    // stricter type-checking for stronger correctness. Recommended by TS
+    "strict": true,
+    // linter checks for common issues
+    "noImplicitReturns": true,
+    "noFallthroughCasesInSwitch": true,
+    // noUnused* overlap with @typescript-eslint/no-unused-vars, can disable if duplicative
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    // use Node's module resolution algorithm, instead of the legacy TS one
```
kanaagawa, line 22: Do you think maybe relevant to link to how the node resolution works: https://www.typescriptlang.org/docs/handbook/module-resolution.html#how-nodejs-resolves-modules

packages/govern-lib/src/utils/NetworkContext.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    // Handle the case of having a name but no chainId.
+    if (name !== undefined && chainId === undefined) {
+      const _chainId = NetworkContext.chainIdFromName(name)
+
+      if (_chainId === undefined) {
+        throw new ErrorInvalidNetwork(
+          `Network: invalid name provided: ${name}. ` +
+            `Please use provide a chainId or use one of the following names: ` +
```
kanaagawa, line 64:
```suggestion
            `Please provide a chainId or use one of the following names: ` +

File: packages/govern-lib/src/utils/NetworkContext.ts [hunk 16]
```diff
@@ -0,0 +1,96 @@
...
+    if (!chainIdNetworkContext) {
+      throw new ErrorInvalidNetwork(
+        `Network: invalid chainId provided: ${chainId}. ` +
+          `Please use one of the following: ` +
+          [...networkContexts.keys()].join(', ') +
+          `.`
+      )
+    }
+
+    if (this.chainId === undefined) {
```
kanaagawa, line 89:
```suggestion
    if (this.chainId === -1) {
```

packages/govern-lib/src/utils/NetworkContext.ts
kanaagawa: I think the network context make sense, specially if we start having more contextual data from places like IPFS.

0xGabi (changes requested):
The changes look good to me. I agree with Sam sentiment about not worry to much about allowing users to integrate with the `core` logic of the server

Issue #54, opened 2020-10-14
Add contract linting


sohkai:
It'd be great to add a contract linting tool to the contract packages. Although not always useful, they can catch some issues or help align code.

Instead of solium, as our old codebases used, I would suggest adding [solhint](https://github.com/protofire/solhint) and using its [prettier extension](https://github.com/fvictorio/solhint-plugin-prettier).

PR #212, opened 2020-11-09
Core: GovernQueue.veto enhancement

### LINKED ISSUE #12: Allow veto on challenged items

nivida:
closes #12 

(This branch is required to finish the test in PR #111)

### DIFF
--- a/packages/govern-core/contracts/pipelines/GovernQueue.sol
+++ b/packages/govern-core/contracts/pipelines/GovernQueue.sol
@@ -227,7 +227,7 @@ contract GovernQueue is IERC3000, AdaptativeERC165, IArbitrable, ACL {
 
             _container.config.challengeDeposit.releaseTo(challengerCache[containerHash]);
             challengerCache[containerHash] = address(0);
-            disputeItemCache[containerHash] = address(0);
+            delete disputeItemCache[containerHash][IArbitrator(_container.config.resolver)];
         } else {
             item.checkAndSetState(
                 GovernQueueStateLib.State.Scheduled,
--- a/packages/govern-core/contracts/pipelines/GovernQueue.sol
+++ b/packages/govern-core/contracts/pipelines/GovernQueue.sol
@@ -217,7 +217,7 @@ contract GovernQueue is IERC3000, AdaptativeERC165, IArbitrable, ACL {
         bytes memory _reason
     ) auth(this.veto.selector) override public {
         bytes32 containerHash = _container.hash();
-        GovernQueueStateLib.Item item = queue[containerHash];
+        GovernQueueStateLib.Item storage item = queue[containerHash];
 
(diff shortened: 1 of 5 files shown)


packages/govern-core/contracts/pipelines/GovernQueue.sol
```diff
@@ -212,13 +212,32 @@ contract GovernQueue is IERC3000, AdaptativeERC165, IArbitrable, ACL {
...
 
-    function veto(bytes32 _containerHash, bytes memory _reason) auth(this.veto.selector) override public {
-        queue[_containerHash].checkAndSetState(
-            GovernQueueStateLib.State.Scheduled,
-            GovernQueueStateLib.State.Cancelled
-        );
+    function veto(
+        ERC3000Data.Container memory _container,
+        bytes memory _reason
+    ) auth(this.veto.selector) override public {
```
Evalir: We should definitely look into #54 so we can have all functions tidy with a certain style. :)

packages/govern-core/contracts/pipelines/GovernQueue.sol
```diff
@@ -212,13 +212,32 @@ contract GovernQueue is IERC3000, AdaptativeERC165, IArbitrable, ACL {
...
+        GovernQueueStateLib.Item storage item = queue[containerHash];
+
+        if (item.state == GovernQueueStateLib.State.Challenged) {
+            item.checkAndSetState(
+                GovernQueueStateLib.State.Challenged,
+                GovernQueueStateLib.State.Cancelled
+            );
+
+            _container.config.challengeDeposit.releaseTo(challengerCache[containerHash]);
+            challengerCache[containerHash] = address(0);
```
Evalir: I think it might be good leaving a comment on why we do this and why it saves gas. :)

packages/govern-core/contracts/pipelines/GovernQueue.sol
```diff
@@ -212,13 +212,32 @@ contract GovernQueue is IERC3000, AdaptativeERC165, IArbitrable, ACL {
...
+            delete disputeItemCache[containerHash][IArbitrator(_container.config.resolver)];
+        } else {
+            item.checkAndSetState(
```
Evalir, line 231: Also, here, it might be good leaving a comment on why do we only explicitly check for only the `Scheduled` state (as all other states are invalid and should revert).

Evalir (approved):
LGTM. 

Let's remember to re-deploy all infra when this is done, as this introduces contract changes! @bpierre @promaty

packages/govern-core/contracts/pipelines/GovernQueue.sol
nivida: I just use the exact same comment from bellow which is I think enough explaining what it is doing.. (it releases the state and refunds the amount of gas cause of it)

PR #213, opened 2020-11-09
Govern E2E: Health check added

### LINKED ISSUE #182: CI/CD: Govern Server E2E test randomly fails
- E2E test run should explicit wait until the server is actually running (flickering behavior currently in the CD run)

nivida:
closes #182

### DIFF
--- a/packages/govern/scripts/start.e2e.sh
+++ b/packages/govern/scripts/start.e2e.sh
@@ -5,7 +5,7 @@ cd ../govern-server/
 nohup yarn dev >>/dev/null 2>>/dev/null &
 
 # Check if the govern-server is running
-while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' localhost:3000/.well-known/apollo/server-health)" != "200" ]]; do sleep 5; done
+yarn server:healthcheck
 
 # Go to govern package and start e2e tests
 cd ../govern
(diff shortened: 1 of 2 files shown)


packages/govern/scripts/start.e2e.sh
```diff
@@ -4,6 +4,9 @@
 cd ../govern-server/
 nohup yarn dev >>/dev/null 2>>/dev/null &
 
+# Check if the govern-server is running
+while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' localhost:3000/.well-known/apollo/server-health)" != "200" ]]; do sleep 5; done
```
promaty: Let's avoid complicated bash scripts whenever we can, as soon as we have `if` / `while` I prefer javascript:
```suggestion
npx wait-on localhost:3000/.well-known/apollo/server-health --timeout 20000
```
Also can you add this into top `package.json` with something like `test:server`

promaty:
Also maybe it's possible to import the server directly into your test? In this way you can see logs from both the server and the test in the same place and you don't need to wait for it to start in a separate process. I did it with express server in protocol for example:
https://github.com/aragon/protocol-backend/blob/development/packages/server/test/integration/user-client-interaction.js#L29

promaty (approved):

PR #214, opened 2020-11-09
Core: Registry proposal

### LINKED ISSUE #196: Core: Registry improvement
The current problem we have with the registry is that we are not able to have multiple pairs stored with only one name. This means to get all the pairs for a more complex DAO is it required to request a list of pairs by their names. This obviously will lead to a bad DX. 

There are two solutions to fix this: 

1. We change the existing mapping and store multiple ``Pair`` metadata structs on-chain and check the DAO owner
2. Instead of storing a struct in the mapping can we just emit a ``Registered`` event and handle the DAO pair set off-chain but do store the owner on-chain to check the permission.

nivida:
I've created quickly a draft implementation of the solution I thought could work for the new ``GovernRegistry``. Would be great to get some feedback here and to start the discussion about.

closes #196

### DIFF
--- a/packages/govern-core/contracts/GovernRegistry.sol
+++ b/packages/govern-core/contracts/GovernRegistry.sol
@@ -13,19 +13,15 @@ import "@aragon/govern-contract-utils/contracts/erc165/ERC165.sol";
 contract GovernRegistry is IERC3000Registry {
     mapping(string => address) public owners;
 
-    modifier onlyOwner(string memory _name) {
-        require(owners[_name] == msg.sender, "registry: only owner");
-        _;
-    }
-
     function register(
         IERC3000Executor _executor,
         IERC3000 _queue,
         IERC20 _token,
         string calldata _name,
         bytes calldata _initialMetadata
-    ) override external onlyOwner(_name)
+    ) override external
     {
+        require(!owners[_name] || owners[_name] == msg.sender, "registry: only owner");
         owners[_name] = msg.sender;
 
         emit Registered(_executor, _queue, _token, msg.sender, _name);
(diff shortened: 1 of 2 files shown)


packages/govern-core/contracts/GovernRegistry.sol
```diff
@@ -36,4 +35,11 @@ contract GovernRegistry is IERC3000Registry {
     function _setMetadata(IERC3000Executor _executor, bytes memory _metadata) internal {
         emit SetMetadata(_executor, _metadata);
```
kanaagawa, line 36: Maybe now the metadata should be related to the owner address instead of the executor?

0xGabi (commented):
My current mental model of Govern is a bit unclear with the current changes. I think we still need to give more definition to what we consider it is a DAO in the context of Govern. 

From the current changes it's clear that: 
1. A DAO will be the owner of several registration entities identified by a unique name. 
2. Those entities are the links between an executor and a queue. 
3. A DAO has metadata and an owner

It's unclear to me:
1. What's the difference between a DAO and an executor instance.
2. What's the relation between register a pair (executor, queue) and assign it an owner.
3. What happened if a pair has no owner.

packages/govern-core/contracts/GovernRegistry.sol
nivida: Yup, agreed. I thought already before it is strange that anyone can change the metadata of a registered ``Executor``.

nivida:
> My current mental model of Govern is a bit unclear with the current changes. I think we still need to give more definition to what we consider it is a DAO in the context of Govern.
> 
> From the current changes it's clear that:
> 
> 1. A DAO will be the owner of several registration entities identified by a unique name.
> 2. Those entities are the links between an executor and a queue.
> 3. A DAO has metadata and an owner
> 
> It's unclear to me:
> 
> 1. What's the difference between a DAO and an executor instance.
> 2. What's the relation between register a pair (executor, queue) and assign it an owner.
> 3. What happened if a pair has no owner.

Some really good points! Do we want to have a call with @izqui to discuss them first before we focus more on the code?

izqui:
@nivida what's the status of this?

nivida:
@izqui We decided back then to organize a call with you to discuss it closer. I will introduce the new devs to Govern and will definitely come back to you in case we have a call about it or any other kind of communication we would like to hear your feedback. Thanks for asking back!

PR #216, opened 2020-11-10
Utils: ACL.bulk test

### LINKED ISSUE #13: Add unit tests for MiniACL.bulk

nivida:
closes #13


0xGabi (approved):
Look good to me!

Maybe we could include one more tests were we bulk a few actions and assert the right number of events are emitted and on the correct order.

0xGabi:
I also notice that lints are not right. Pierre include a git hook with husky for the `GovernLib` maybe we should standardize that for every package on the monorepo. cc @bpierre

PR #217, opened 2020-11-10
Utils: ERC1167ProxyFactory test

### LINKED ISSUE #119: Utils: Write test for ERC1167ProxyFactory library

nivida:
closes #119


packages/govern-contract-utils/test/erc1167-proxy-factory.test.ts
```diff
@@ -0,0 +1,120 @@
+import { expect } from 'chai'
+import { ethers } from 'hardhat'
+import { Signer } from 'ethers'
+import { defaultAbiCoder } from 'ethers/lib/utils'
+import { generateCodeFromContract } from './helpers'
```
kanaagawa: I think we can remove this line.

packages/govern-contract-utils/test/erc1167-proxy-factory.test.ts
```diff
@@ -0,0 +1,120 @@
...
+import {
+  CloneFactoryMockFactory,
+  CloneFactoryMock,
+  ClonedContract,
+  ClonedContractWithInit
+} from '../typechain'
+import
+  * as ClonedContractArtifact
```
kanaagawa: Maybe instead of importing the contract artifacts this way to get the abi. We can use the contract names, I think `ethers.getContractAt` accept the contract name as well.

packages/govern-contract-utils/test/erc1167-proxy-factory.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    owner: string,
+    factory: CloneFactoryMock,
+    clonedContractAddress: string,
+    clonedContract: any
+
+  before(async () => {
+    signers = await ethers.getSigners()
+    owner = await signers[0].getAddress()
+    const CloneFactoryMock = (await ethers.getContractFactory('CloneFactoryMock')) as CloneFactoryMockFactory
+    factory = await CloneFactoryMock.deploy()
```
kanaagawa, line 24: This seems a common patter across tests. I already did a few helper on the `govern-create` package that we can reuse here. I wonder if maybe the best place is to have them in this repo (`govern-contract-utils`) and export them in other places 🤔

packages/govern-contract-utils/test/erc1167-proxy-factory.test.ts
```diff
@@ -0,0 +1,120 @@
...
+  context('Clone', () => {
+    it('clone without constructor parameters', async () => {
+      await factory.clone()
+
+      clonedContractAddress = await factory.latestClonedContract()
+      clonedContract = (await ethers.getContractAt(
+        ClonedContractArtifact.abi,
+        clonedContractAddress,
+        signers[0]
+      )) as ClonedContract
```
kanaagawa: Similar comment that above, we can reuse a helper to accept types like: `getContractAt<T>`

packages/govern-contract-utils/test/erc1167-proxy-factory.test.ts
```diff
@@ -0,0 +1,120 @@
...
+        .to.be.equal('INIT DATA')
+    })
+  })
+
+  context('Helper methods', () => {
+    it('calls "generateCode" as expected', async () => {
+      await factory.generateCode()
+
+      expect(await factory.generatedCode())
+        .to.equal('0x3d602d80600a3d3981f3363d3d373d3d3d363d739467a509da43cb50eb332187602534991be1fea45af43d82803e903d91602b57fd5bf3')
```
kanaagawa, line 92: Did we get this value from the artifact? Or check manually the generate code?

packages/govern-contract-utils/contracts/test/CloneFactoryMock.sol
```diff
@@ -0,0 +1,68 @@
+/*
```
kanaagawa, line 1: I think maybe interesting to give this plugin a try: https://hardhat.org/plugins/eth-optimisim-smock.html#mockcontractfunction

I'll try to use it on my tests and let you know 👍

0xGabi (approved):
Great work! 💪 I left a few comments, mostly cosmetics about how we structure the code.

packages/govern-contract-utils/contracts/test/CloneFactoryMock.sol
nivida: Yes, this is definitely something we can use in some cases. This specific case doesn't include it :-)

packages/govern-contract-utils/test/erc1167-proxy-factory.test.ts
nivida: I decided to not implement the ``generateCode`` logic in JS to lower the complexity.. but I'm open to doing it.. The thing is we would just test the solidity language itself in the end.. This means the test helpers I wrote for ``ERC3000Data`` for example are also not really required but I felt like doing them as a short refresher after the seven months sabbatical..

packages/govern-contract-utils/test/erc1167-proxy-factory.test.ts
nivida: Feel free to create a related issue and we can do this in another PR and apply it on all tests.. I think it is better to keep the scope of a PR as small as possible otherwise some mistakes could get merged..

packages/govern-contract-utils/test/erc1167-proxy-factory.test.ts
nivida: Thanks for the hint.. will check that closer 👍

packages/govern-contract-utils/test/erc1167-proxy-factory.test.ts
nivida: > Or check manually the generate code?

... and to answer your question:

I didn't have to check the returned binary of ``generateCode`` itself cause the above test cases do already assert that the generated code does work as expected and ``clone`` does deploy a proxy contract with his own state. This means if ``generateCode`` doesn't work properly do also the test cases above not work because it is already tested indirectly.

kanaagawa: Great, thanks for the explanation.

PR #223, opened 2020-11-13
Utils: add adaptiveERC165 test

### LINKED ISSUE #113: Utils: Write test for AdaptativeERC165

0xGabi:
Close #113


packages/govern-core/test/govern.test.ts
```diff
@@ -84,21 +84,12 @@ describe('Govern', function() {
   })
 
   context('ERC-165', () => {
-    const ERC165_INTERFACE_ID = '0x01ffc9a7'
```
nivida, line 95: We can keep those cases here just to assert ``Govern`` is providing the ERC165 methods. But we could for example just check the generated ABI instead of doing actual requests.

packages/govern-contract-utils/test/adaptative-erc165.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  REGISTERED_CALLBACK: 'RegisteredCallback',
+  REGISTERED_STANDARD: 'RegisteredStandard',
+  RECEIVED_CALLBACK: 'ReceivedCallback',
+}
+
+const beefInterfaceId = '0xbeefbeef'
+const callbackSig = hexDataSlice(id('callbackFunc()'), 0, 4) // 0x1eb2075a
+const magicNumber = '0x10000000'
+
+// NOTE: We cannot assert on the values emitted
```
nivida: I think we can remove this note because it is possible in this test to assert the values of the events as expected. Only structs who do include the ``BigNumber`` object can't be asserted correctly and it doesn't depend if they are in an event or not. As soon as a ``BigNumber`` object is nested in an ``Array`` is this bug existing.

packages/govern-contract-utils/test/adaptative-erc165.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        .withArgs(callbackSig, magicNumber)
+    })
+
+    it('ensures support the right interfaceID', async () => {
+      await expect(
+        await adaptative.supportsInterface(beefInterfaceId)
+      ).to.equal(true)
+    })
+
+    it('ensures the right callback was call', async () => {
```
nivida: We should here also test the inline assembly of the ``_handleCallback`` method. It does return the memory value at position ``0x00`` with an offset of ``0x20``.

packages/govern-contract-utils/package.json
```diff
@@ -22,11 +22,11 @@
     "hardhat": "^2.0.2"
   },
   "scripts": {
-    "build": "yarn compile && hardhat typechain",
+    "build": "yarn compile",
```
nivida: What exactly was the reason to change those commands within this PR?

nivida (changes requested):
As I was seeing are the ``govern-contract-utils`` test cases not running in the CI.

packages/govern-core/test/govern.test.ts
nivida: I anyways missing the ``supportsInterface`` method in the ``Govern`` contract. I will check that closer asap.

packages/govern-contract-utils/package.json
kanaagawa: The command hardhat type chain is not longer relevant as types are generated during compilation.

packages/govern-contract-utils/test/adaptative-erc165.test.ts
kanaagawa: I'm not sure how to test this. How do you suggest we should do it?

nivida:
@0xGabi What is the status of this PR? have you extended the coverage as requested and discussed on discord?

izqui:
@nivida can you finish this PR and merge it?

nivida (approved):

PR #237, opened 2020-12-03
Console: replace Apollo Client with React Query

Evalir:
The rationale for this change is mainly due to how heavy / unflexible Apollo client can be when thinking outside a GraphQL context. As we might be integrating with external APIs for DAO Activity / Snapshot, it makes sense to move to something more protocol agnostic, but that still provides all the nice things, such as retries and caching. react-query has become a community standard and is well maintained, so, in my opinion, it's a good choice.


packages/govern-console/src/lib/utils.ts
```diff
@@ -1,3 +1,6 @@
+export const MAINNET_SUBGRAPH_URL = 'https://api.thegraph.com/subgraphs/name/aragon/aragon-govern-mainnet'
+export const RINKEBY_SUBGRAPH_URL = 'https://api.thegraph.com/subgraphs/name/evalir/aragon-govern-rinkeby'
```
bpierre: Should we move these to a file like `env` on `consts`?

packages/govern-console/src/pages/ViewDao.tsx
```diff
@@ -86,11 +87,16 @@ export default function DaoView() {
...
+    const res = await request(
+      chainId === 4 ? RINKEBY_SUBGRAPH_URL : MAINNET_SUBGRAPH_URL,
+      DAO_QUERY,
+      {
+        name: daoAddress,
+      },
+    )
+
+    return res
   })
```
bpierre: I think we could do this directly since `request()` returns a promise?

```suggestion
  const { data, isLoading: loading, error } = useQuery('DAO_DATA', () => request(
    chainId === 4 ? RINKEBY_SUBGRAPH_URL : MAINNET_SUBGRAPH_URL,
    DAO_QUERY,
    { name: daoAddress },
  ))
```

bpierre (approved):
:100: 

The bundle size must be much smaller now!

packages/govern-console/src/pages/ViewDao.tsx
Evalir: Yes!

packages/govern-console/src/lib/utils.ts
Evalir: Yes, let's do it!

EDIT: Done with https://github.com/aragon/govern/pull/237/commits/fcc54198aef8f49a036fe2dbfaf2df944bb87c93

PR #238, opened 2020-12-03
CI: cache abi files

0xGabi:


.github/workflows/subgraph-ci-cd.yml
```diff
@@ -24,10 +24,20 @@ jobs:
       with:
         path: '**/node_modules'
         key: yarn-${{ hashFiles('yarn.lock') }}
+    
+    - name: Cache abi
+      uses: actions/cache@v2
+      with:
+        path: |
+          packages/*/abi
```
promaty: This doesn't help, because contracts will still be compiled if there is no `artifacts` or `cache` present.

Also I don't think that extra plugins care about existing files, they always run on top (like the typechain plugin).

So if you want to speed this part up, just copy contract caching from contracts-ci-cd

promaty (changes requested):

.github/workflows/subgraph-ci-cd.yml
kanaagawa: Make sense, thanks for the feedback!

PR #245, opened 2020-12-07
Console: base routing on network / chain ID rather than DAO address

### LINKED ISSUE #241: Console: move to generic routing based on chain ID
Right now, we're using a dropdown switch for changing the network we're connected to. While this is "okay", it doesn't let users share links to their DAO. A more generic routing system which adds the chain to the route let's us share things like `console.aragon.org/mainnet/bitconnect` or `console.aragon.org/rinkeby/rocketpool` with much more ease.

Evalir:
Closes #241.

- Moves routing to have a parameter to track the network (`mainnet` or `rinkeby`)
- Introduces a redirect to the DAO Selector when the network is changed on the dropdown
- Introduces a home button (the logo on the top left, just like client.aragon.org)

_pls ignore that accidental push to master :D_

### DIFF
--- a/packages/govern-console/src/components/Header/Header.tsx
+++ b/packages/govern-console/src/components/Header/Header.tsx
@@ -1,4 +1,5 @@
 import React, { useCallback, useEffect } from 'react'
+import { useHistory } from 'react-router-dom'
 import { ChainUnsupportedError } from 'use-wallet'
 import 'styled-components/macro'
 import Button from '../Button'
@@ -10,14 +11,7 @@ import AragonSvg from '../../assets/aragon-metal.svg'
 function Header() {
   const { chainId, setChainId } = useChainId()
   const { wallet } = useWallet()
-
-  const handleWalletConnection = useCallback(() => {
-    wallet.status === 'connected' ? wallet.reset() : wallet.connect('injected')
-  }, [wallet])
-
-  const handleChangeChain = useCallback(e => setChainId(e.target.value), [
-    setChainId,
-  ])
+  const history = useHistory()
 
   useEffect(() => {
     if (wallet!.error && wallet!.error instanceof ChainUnsupportedError) {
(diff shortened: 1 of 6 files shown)


packages/govern-console/src/components/Header/Header.tsx
```diff
@@ -47,7 +56,23 @@ function Header() {
...
+            &:hover {
+              cursor: pointer;
+            }
+            &:active {
+              position: relative;
+              top: 1px;
+            }
+          `}
+        >
+          <img src={AragonSvg} width="36" alt="Grey Eagle Aragon logo" />
```
bpierre: We should use an empty alt here, since we have the title right after it (“Govern Console”). When a logo is being used in a header, it is generally recommended to use either an empty alt (when the textual version is present) or the name of the site / app directly, rather than describing the image itself. This is to avoid screen readers repeating “Grey Eagle Aragon logo − Govern Console” every time the page loads :smile: 

Now since we are adding a button here, we need a way to describe it. We could use a `title` to describe the action, but should we move the `h1` inside the button instead? I think it would make sense to make them both clickable.

bpierre (approved):
LGTM! Just left a comment about the `alt=""` attribute for the logo.

packages/govern-console/src/components/Header/Header.tsx
Evalir: I had no idea about this for screen readers! Will do. :) Do you have any material regarding this that I could read? And yes, let's actually move it inside the button. Looking back, it feels weird that you can click the logo but not the title. 😄 

Done in https://github.com/aragon/govern/pull/245/commits/05c49a7dafb23387391101e507c48fb78b275a35

bpierre: > Do you have any material regarding this that I could read?

Of course!

The WAI tutorial about images describes a case similar to this one: https://www.w3.org/WAI/tutorials/images/functional/#logo-image-within-link-text

Except in our case, the text doesn’t describe the actual action… but a click on a logo is generally expected to have the “go to home” behavior, so I usually omit it. This is also [what the WebAIM website recommends](https://webaim.org/techniques/alttext/#logos):

> Developers will often identify the image as being to the home page (alt="Acme Company home page), though if the image is consistently at the beginning of the page and the alternative text is appropriate, this additional information should not be needed.

PR #247, opened 2020-12-08
govern-contract-utils: optimize ERC165 by introducing compiler visibility warning

sohkai:
Debatable but created to gauge how we feel.

Introducing this commit will introduce compiler warnings about the `ERC165#supportsInterface()` function visibility not being pure, but it does seem awkward to waste gas just to satisfy this compiler warning. This view function is intended to be used on-chain, so it'd be a very small gas overhead.


sohkai:
Will save this; perhaps upgrading to solc@0.7.5 solves the compiler warning.

nivida (approved):
The changes look good to me.. and yes let us see if a solc upgrade does solve the warnings 👍

izqui (approved):

PR #248, opened 2020-12-08
govern-core, govern-contract-utils: callback magic numbers should be bytes32

sohkai:
I'm pretty sure this was a typo on copying the function selectors; callback magic numbers are usually requiring the full `bytes32`.


sohkai:
Oops, just checked some ERCs and it turns out the magic numbers are usually of `bytes4`. Not sure where I got the misconception of requiring `bytes32`!

izqui:
It might actually be safer to allow bytes32 for future proofing

PR #249, opened 2020-12-08
govern-contract-utils: allow ACLOracle check to fail and push ACL#willPerform()'s duplicated role checks into isolated function

sohkai:
Previously, the ACLOracle check could revert and cause a user with a reverting ACLOracle defined for their address to not be allowed even though any account was allowed (although it could be said that this particular ACL state would be considered a user configuration mistake in and of itself). Allowing the ACLOracle check to gracefully fail costs a bit more gas, but covers this case.

Pushing the role check into its own function makes the code a bit more DRY and simpler, albeit at a slight gas overhead.


packages/govern-contract-utils/contracts/acl/ACL.sol
```diff
@@ -130,7 +119,20 @@ contract ACL is Initializable {
...
         emit Frozen(_role, msg.sender);
     }
+
+    function _checkRole(bytes4 _role, address _who, bytes memory _data) internal returns (bool) {
+        address accessFlagOrAclOracle = roles[_role][_who];
+        if (accessFlagOrAclOracle != UNSET_ROLE) {
+            if (accessFlagOrAclOracle == ALLOW_FLAG) return true;
+
+            // Since it's not a flag, assume it's an ACLOracle and try-catch to skip failures
+            try IACLOracle(accessFlagOrAclOracle).willPerform(_role, _who, _data) returns (bool allowed) {
```
izqui, line 131: This is really nice, first time I see this

izqui:
lgtm

PR #250, opened 2020-12-08
govern-contract-utils: ensure ACL is initialized with correct root account

sohkai:
Also debatable, but I found it weird that the initialization check would just get skipped if another constructor/function had already initialized the ACL. Particularly, I was afraid of a contract being statically or dynamically misconfigured such that two initialization areas contained different root grantees.

### DIFF
--- a/packages/govern-contract-utils/contracts/acl/ACL.sol
+++ b/packages/govern-contract-utils/contracts/acl/ACL.sol
@@ -48,6 +48,8 @@ contract ACL is Initializable {
         // ACL might have been already initialized by constructors
         if (initBlocks["acl"] == 0) {
             _initializeACL(_initialRoot);
+        } else {
+            require(roles[ROOT_ROLE][_initialRoot] == ALLOW_FLAG, "acl: initial root misaligned");
         }
         _;
     }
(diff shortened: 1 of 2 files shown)


sohkai:
Good to know my suspicion wasn't misaligned; adding this creates a slight 1k gas overhead but safeguards us from mistakes like the one found in fe129f7 :).

packages/govern-core/contracts/Govern.sol
```diff
@@ -21,7 +21,7 @@ contract Govern is AdaptativeERC165, IERC3000Executor, ACL {
 
     event ETHDeposited(address indexed sender, uint256 value);
 
-    constructor(address _initialExecutor) ACL(address(this)) public {
+    constructor(address _initialExecutor) ACL(address(_initialExecutor)) public {
```
izqui, line 24: This was actually intentional. The root of Govern should be Govern itself.

This means that Govern itself (via action execution) can change the ACL in order to add other executors. The initial executor should have execution granted but it shouldn't be root IMO

packages/govern-contract-utils/contracts/acl/ACL.sol
```diff
@@ -48,6 +48,8 @@ contract ACL is Initializable {
         // ACL might have been already initialized by constructors
         if (initBlocks["acl"] == 0) {
             _initializeACL(_initialRoot);
+        } else {
+            require(roles[ROOT_ROLE][_initialRoot] == ALLOW_FLAG, "acl: initial root misaligned");
```
izqui, line 52: Nice one

packages/govern-core/contracts/Govern.sol
sohkai: See https://github.com/aragon/govern/pull/257

PR #251, opened 2020-12-08
govern-contract-utils: lightly update DepositLib's event and error API

sohkai:
Generalizes `DepositLib` into the utils, and updates its events to fit the past-tense nomenclature of the other events.

Wasn't sure if we depended on this event anywhere else; as far as I could search there was nothing in a frontend, API, or subgraph package using these events or errors.


izqui (commented):
lgtm

PR #252, opened 2020-12-08
govern-contract-utils: fix ERC1167ProxyFactory#generateCode()

### LINKED ISSUE #239: Create: Factories not working correctly
It seems that the transparent proxies implemented for the factories (#88) have introduced a very cryptic bug somewhere; I haven't found the case, but I do know that if we entirely remove the proxies and just use "vanilla" factories, contracts work as expected. [Here](https://rinkeby.etherscan.io/tx/0x654e8c526120b6619673729151ec4286ee837de91c4a244bdbf96617921a7376)'s an example of a TX to a GovernFactory to create a new DAO.

sohkai:
Fixes #239 (I think).

Updates how the data given to the `CREATE` opcode is generated, and verified output with the EIP in tests.

### DIFF
--- a/packages/govern-contract-utils/contracts/minimal-proxies/ERC1167ProxyFactory.sol
+++ b/packages/govern-contract-utils/contracts/minimal-proxies/ERC1167ProxyFactory.sol
@@ -8,10 +8,10 @@ pragma solidity ^0.6.8;
 
 library ERC1167ProxyFactory {
     function clone(address _implementation) internal returns (address cloneAddr) {
-        bytes memory code = generateCode(_implementation);
+        bytes memory createData = generateCreateData(_implementation);
 
         assembly {
-            cloneAddr := create(0, add(code, 0x20), 55)
+            cloneAddr := create(0, add(createData, 0x20), 55)
         }
         
         require(cloneAddr != address(0), "proxy-factory: bad create");
@@ -25,10 +25,10 @@ library ERC1167ProxyFactory {
     }
 
     function clone2(address _implementation, bytes32 _salt) internal returns (address cloneAddr) {
-        bytes memory code = generateCode(_implementation);
-        
+        bytes memory createData = generateCreateData(_implementation);
+
         assembly {
-            cloneAddr := create2(0, add(code, 0x20), 55, _salt)
+            cloneAddr := create2(0, add(createData, 0x20), 55, _salt)
         }
         
         require(cloneAddr != address(0), "proxy-factory: bad create2");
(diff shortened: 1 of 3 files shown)


packages/govern-contract-utils/contracts/minimal-proxies/ERC1167ProxyFactory.sol
```diff
@@ -41,14 +41,15 @@ library ERC1167ProxyFactory {
...
 
-    function generateCode(address _implementation) internal pure returns (bytes memory code) {
-        code = new bytes(55);
-        
-        assembly {
-            mstore(add(code, 0x20), 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
-            mstore(add(code, 0x34), shl(0x60, _implementation))
-            mstore(add(code, 0x48), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
-        }
+    function generateCreateData(address _implementation) internal pure returns (bytes memory) {
```
sohkai, line 44: Updated the name as it's not really generating code (that would be the bit after the "proxy code" comment). This data includes the constructor as well.

packages/govern-contract-utils/contracts/minimal-proxies/ERC1167ProxyFactory.sol
```diff
@@ -41,14 +41,15 @@ library ERC1167ProxyFactory {
...
+        return abi.encodePacked(
```
sohkai, line 45: I think this is much cleaner than `mstore`s in assembly, but on a gas check, it also seems to shave one or two gas as well :smile:

izqui: This is so much better. I love using this technique for string concat as well

packages/govern-contract-utils/contracts/minimal-proxies/ERC1167ProxyFactory.sol
izqui: Fair

packages/govern-contract-utils/contracts/minimal-proxies/ERC1167ProxyFactory.sol
```diff
@@ -41,14 +41,15 @@ library ERC1167ProxyFactory {
...
+            //---- constructor -----
+            bytes10(0x3d602d80600a3d3981f3),
+            //---- proxy code -----
+            bytes10(0x363d3d373d3d3d363d73),
+            _implementation,
+            bytes15(0x5af43d82803e903d91602b57fd5bf3)
```
izqui, line 51: Does the new code return the error data on revert?

izqui (commented):
Still testing, unsure of why it is failing now:

Screen Shot 2020-12-08 at 9 02 56 AM

Edit: it seems like there were some issues deploying the Factory, will update later

packages/govern-contract-utils/contracts/minimal-proxies/ERC1167ProxyFactory.sol
sohkai: This is the same code in EIP-1167; from the opcode table on the EIP page, I think it should be.

izqui: Should we make the tests for Govern.sol and GovernQueue.sol run both against the contract directly and the contract behind a proxy as we used to do for some contracts aragonOS and/or aragon-apps?

nivida: We currently only have unit tests and could extend our contract test suite to include integration tests for the whole govern contract framework. This would automatically include the mentioned additional tests from @izqui. A test suite like this would give us the highest possible proof that anything is working as expected.

PR #253, opened 2020-12-08
erc3k: use abi.encode() for hashing payload data

sohkai:
Aligns the ABI encoding of `ERC3000Data`'s payload and config hashing.

These were misaligned in that hashing the payload struct used `abi.encodePacked()` whereas hashing the config struct used `abi.encode()`. I didn't see much reason to use `encodePacked()` for the payload (and for encoding structs, using the "standard" ABI encoding is generally preferred from what I've seen), but let me know otherwise.


izqui (approved):
Can't remember why I decided to used `encodePacked`

PR #255, opened 2020-12-08
Create: refactor to use hardhat with hardhat-deploy

### LINKED ISSUE #175: Create: refactor

0xGabi:
Close #175 

The changes in this PR include:
- Migrate package to typescript + hardhat
- Use `hardhat-deploy` as deployment tool
- Etherscan verification


packages/govern-create/deploy/factories.ts
```diff
@@ -0,0 +1,48 @@
+import { HardhatRuntimeEnvironment } from 'hardhat/types'
+import { DeployFunction } from 'hardhat-deploy/types'
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
```
nivida, line 4: We probably can find a better name for this function and I think it is not required to assign it to a const :)

packages/govern-create/deploy/factories.ts
```diff
@@ -0,0 +1,48 @@
...
+  const { deployments, getNamedAccounts } = hre
+  const { deploy } = deployments
+
+  const { deployer } = await getNamedAccounts()
```
nivida, line 8: It would be possible to type those const's as well.

packages/govern-create/deploy/factories.ts
```diff
@@ -0,0 +1,48 @@
...
+    args: [
+      registry.address,
+      governFactory.address,
+      queueFactory.address,
+      tokenFactory.address,
+    ],
+    log: true,
+  })
+}
+export default func
```
nivida, line 41: Can we have the export as the last statement?

packages/govern-create/deploy/registry.ts
```diff
@@ -0,0 +1,17 @@
+import { HardhatRuntimeEnvironment } from 'hardhat/types'
+import { DeployFunction } from 'hardhat-deploy/types'
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
```
nivida, line 4: Same comments as in ``factories.ts``

packages/govern-create/deploy/verify.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const dummyConfig = {
+  executionDelay: '0',
+  scheduleDeposit: [NO_TOKEN, '0'],
+  challengeDeposit: [NO_TOKEN, '0'],
+  resolver: ZERO_ADDR,
+  rules: '0x',
+}
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
```
nivida, line 19: Same comments as on the other files.

packages/govern-create/deploy/verify.ts
```diff
@@ -0,0 +1,63 @@
...
+  const { deployments, ethers, run } = hre
+
+  console.log('Verifying registry and factories contracts')
+
+  await run(TASK_ETHERSCAN_VERIFY, {
+    apiKey: process.env.ETHERSCAN_KEY,
+    license: 'GPL-3.0',
```
nivida, line 26: Can we probably import the license from the ``package.json``?

packages/govern-create/hardhat.config.ts
```diff
@@ -0,0 +1,57 @@
+import { config as dotenvConfig } from 'dotenv'
+import { resolve } from 'path'
+dotenvConfig({ path: resolve(__dirname, './.env') })
```
nivida, line 3: Is it required to do this call on this line? or can we have it after all imports to increase the code style? :)

packages/govern-create/tasks/govern.ts
```diff
@@ -0,0 +1,87 @@
...
+    uniqueNamesGenerator({
+      dictionaries: [adjectives, colors, animals],
+      length: 2,
+      separator: '-',
+    })
+
+  return process.env.CD ? `github-${uniqueName}` : uniqueName
+}
+
+const format = (address: string, name: string, network: string): string =>
```
nivida, line 22: Can we have one type of defining functions? I would prefer to use the normal function keyword for many reasons.

packages/govern-create/tasks/govern.ts
```diff
@@ -0,0 +1,87 @@
...
+  `- ${name}: https://${
+    network === 'mainnet' ? '' : `${network}.`
+  }etherscan.io/address/${address}`
+
+task('deploy-govern', 'Deploys a Govern instance')
+  .addOptionalParam('factory', 'GovernBaseFactory address')
+  .addOptionalParam('useProxies', 'Whether to deploy govern with proxies')
+  .addOptionalParam('name', 'DAO name (must be unique at GovernRegistry level)')
+  .setAction(
+    async (
```
nivida, line 32: We could have this also as a non-anonymous function and with a good name as the others. Would increase the readability.

packages/govern-create/tasks/govern.ts
```diff
@@ -0,0 +1,87 @@
...
+        token = `0x${'00'.repeat(20)}`,
+        tokenName = name,
+        tokenSymbol = 'GOV',
+      },
+      HRE
+    ) => {
+      name = buildName(name)
+      const { ethers, deployments, network } = HRE
+
+      const registryDeployment = await deployments.get('GovernRegistry')
```
nivida, line 46: Also here can we type way more.

packages/govern-create/tasks/govern.ts
```diff
@@ -0,0 +1,87 @@
...
+
+      const args = logs
+        .filter(({ address }) => address === registryContract.address)
+        .map((log) => registryContract.interface.parseLog(log))
+        .find(({ name }) => name === 'Registered')
+
+      const queueAddress = args?.args[1] as string
+      const governAddress = args?.args[0] as string
+
+      if (network.name !== 'hardhat' && network.name !== 'coverage') {
```
nivida, line 81: ``if (network.name !== 'hardhat' || network.name !== 'coverage')``

packages/govern-create/test/base.ts
```diff
@@ -0,0 +1,63 @@
...
+      useProxies
+    )
+
+    await expect(tx).to.emit(registryContract, EVENTS.REGISTERED)
+    await expect(tx).to.emit(registryContract, EVENTS.SET_METADATA)
+
+    const { hash } = await tx
+    const { gasUsed } = await waffle.provider.getTransactionReceipt(hash)
+
+    expect(gasUsed).to.be.lte(gasTarget)
```
nivida, line 46: Wouldn't it be lower than or equals?

packages/govern-create/test/base.ts
```diff
@@ -0,0 +1,63 @@
+import { deployments, ethers, network, waffle } from 'hardhat'
```
nivida, line 1: Is it possible to name the file as the contract is named we do test here? (GovernBaseFactoryTest)

packages/govern-create/test/base.ts
```diff
@@ -0,0 +1,63 @@
...
+    await deployDAO(false, GAS_TARGET)
+  })
+
+  const GAS_TARGET_PROXY = network.name !== 'coverage' ? 6e5 : 2e6
+  it(`deploys DAO with proxies under ${GAS_TARGET_PROXY} gas`, async () => {
+    await deployDAO(true, GAS_TARGET_PROXY)
+  })
+
+  // TODO: Implement tests
+  it('deploys DAO with token')
```
nivida, line 62: Have you thought to implement this test before the merge? Otherwise, add a skip.

packages/govern-create/utils/etherscan.ts
```diff
@@ -0,0 +1,77 @@
+import HRE from 'hardhat'
+import fs from 'fs'
+import { file } from 'tmp-promise'
+
+export const SUPPORTED_ETHERSCAN_NETWORKS = ['mainnet', 'rinkeby']
+
+function delay(ms: number) {
```
nivida, line 7: Missing type for the return value

packages/govern-create/utils/etherscan.ts
```diff
@@ -0,0 +1,77 @@
+import HRE from 'hardhat'
+import fs from 'fs'
+import { file } from 'tmp-promise'
+
+export const SUPPORTED_ETHERSCAN_NETWORKS = ['mainnet', 'rinkeby']
```
nivida, line 5: Missing type. I think we can here anywhere more types. :-)

packages/govern-create/utils/etherscan.ts
```diff
@@ -0,0 +1,77 @@
...
+export const verifyContract = async (
+  address: string,
+  constructorArguments: any[]
+) => {
+  const currentNetwork = HRE.network.name
+
+  if (!process.env.ETHERSCAN_KEY) {
+    throw Error('Missing process.env.ETHERSCAN_KEY.')
+  }
+  if (!SUPPORTED_ETHERSCAN_NETWORKS.includes(currentNetwork)) {
```
nivida, line 20: Missing break

packages/govern-create/utils/etherscan.ts
```diff
@@ -0,0 +1,77 @@
...
+      fd,
+      `module.exports = ${JSON.stringify([...constructorArguments])};`
+    )
+
+    const params = {
+      address: address,
+      constructorArgs: path,
+    }
+    await runTaskWithRetry('verify', params, times, msDelay, cleanup)
+  } catch (error) {}
```
nivida, line 47: What is the rationale to hide this error?

packages/govern-create/utils/etherscan.ts
```diff
@@ -0,0 +1,77 @@
...
+  params: any,
+  times: number,
+  msDelay: number,
+  cleanup: () => void
+) => {
+  let counter = times
+  await delay(msDelay)
+
+  try {
+    if (times) {
```
nivida, line 61: Not sure if the retry mechanism is actually working as expected cause of the way the counter is handled. Shouldn't be ``times--`` on line 71?

packages/govern-create/utils/network.ts
```diff
@@ -0,0 +1,43 @@
+import 'dotenv/config'
+export function node_url(networkName: string): string {
```
nivida, line 2: Missing break

packages/govern-create/utils/network.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (uri) {
+    uri = uri.replace('{{networkName}}', networkName)
+  }
+  if (!uri || uri === '') {
+    // throw new Error(`environment variable "ETH_NODE_URI" not configured `);
+    return ''
+  }
+  if (uri.indexOf('{{') >= 0) {
+    throw new Error(
+      `invalid uri or network not supported by nod eprovider : ${uri}`
```
nivida, line 20: Typo: ``...by nod eprovider...``

packages/govern-create/utils/network.ts
```diff
@@ -0,0 +1,43 @@
...
+      return uri
+    }
+  }
+
+  let uri = process.env.ETH_NODE_URI
+  if (uri) {
+    uri = uri.replace('{{networkName}}', networkName)
+  }
+  if (!uri || uri === '') {
+    // throw new Error(`environment variable "ETH_NODE_URI" not configured `);
```
nivida, line 15: What was the reason to keep this out-commented?

packages/govern-create/utils/network.ts
nivida, line 2: Can you elaborate the logic of this closer to me? not directly clear to me with just reading the code.

packages/govern-create/utils/network.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  const mnemonic = process.env.MNEMONIC
+  if (!mnemonic || mnemonic === '') {
+    return 'test test test test test test test test test test test junk'
+  }
+  return mnemonic
+}
+
+export function accounts(networkName?: string): { mnemonic: string } {
```
nivida, line 41: Here we use the function keyword as I prefer and on other files not (e.g.: ``etherscan.ts``).. is there a reason for?

packages/govern-create/utils/network.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (networkName) {
+    const mnemonic = process.env['MNEMONIC_' + networkName.toUpperCase()]
+    if (mnemonic && mnemonic !== '') {
+      return mnemonic
+    }
+  }
+
+  const mnemonic = process.env.MNEMONIC
+  if (!mnemonic || mnemonic === '') {
+    return 'test test test test test test test test test test test junk'
```
nivida, line 36: Is it really thought to return this string?

nivida (changes requested):
First of all sorry for all those comments. But I think most of the comments are just cosmetics I have seen. By side of those would I like to ask you if you have thought to abstract the deployment logic with 1-2 simple objects to increase the readability and extensibility of it? :)

Edit: I think we already had a similar PR where I have commented on a lot of questions. At least the code looks similar on many parts.

packages/govern-create/test/base.ts
```diff
@@ -0,0 +1,63 @@
...
+  beforeEach(async () => {
+    await deployments.fixture()
+  })
+
+  const deployDAO = async (
+    useProxies: boolean,
+    gasTarget: number,
+    deployToken = false
+  ) => {
+    const registryDeployment = await deployments.get('GovernRegistry')
```
nivida, line 20: What was the idea behind moving the base test setup to this function instead of keeping it in the beforeEach call?

packages/govern-create/deploy/factories.ts
kanaagawa: I agree, but here types are already inferred. What you consider are the benefits of adding type definitions anyways?

packages/govern-create/deploy/factories.ts
kanaagawa: This was the suggested way to write hardhat-deploy scripts per their docs. I'll choose a more meaninful function name. About `const` I think assign in this way make it clear the type we are using: `DeployFunction`.

packages/govern-create/deploy/verify.ts
kanaagawa: Good point!

packages/govern-create/tasks/govern.ts
kanaagawa: This is the usual way to write hardhat tasks. The name of the task is already stated (`deploy-govern`). Here we are just setting the action to be done. Why you consider it's not readable.

packages/govern-create/tasks/govern.ts
kanaagawa: The logic I want to state is only print when the network name is not `hardhat` and is not `coverage`. I think the one you share would always be true.

packages/govern-create/test/base.ts
kanaagawa: As I understood this is what `lte` is doing: https://www.chaijs.com/api/bdd/#method_most

packages/govern-create/deploy/factories.ts
nivida: > This was the suggested way to write hardhat-deploy scripts per their docs. I'll choose a more meaninful function name.

Okay, I will ask Pato why exactly they document it like this. 

>  About const I think assign in this way make it clear the type we are using: DeployFunction.

This is also easily possible with ``function test(...): DeployFunction {}``

packages/govern-create/deploy/factories.ts
nivida: To have the full type security because the current definition is an implicit any type definition and wouldn't provide the full type security you could have with TS.

packages/govern-create/tasks/govern.ts
nivida: Was just a spontaneous thought I had. We can keep it as it is. I myself would just have registered the functions here and would have had the explicit logic somewhere else. This because of the "separation of concerns" thought. 

> Why you consider it's not readable.

It is readable but we could increase the readability a bit :-) 

Marked as resolved because you just follow some other patterns than I do.

packages/govern-create/tasks/govern.ts
nivida: Yup, you are right. Sorry was a brainfart.

packages/govern-create/test/base.ts
nivida: Sorry I missed the ``e`` in ``lte``

packages/govern-create/deploy/factories.ts
nivida: nvm.. Didn't knew the ``DeployFunction`` type is actually a hardhat function type. I thought it will return such a function.

packages/govern-create/test/base.ts
kanaagawa: You are right, we should keep it in the `beforeEach`

packages/govern-create/utils/etherscan.ts
kanaagawa: This was a snipped of code I forked from other project. As I understood the code was to prevent the hardhat `script/task/deploy` to stop execution because this error.

packages/govern-create/utils/etherscan.ts
kanaagawa: Actually in this line we setup `counter` with `times` value: https://github.com/aragon/govern/pull/255/files#diff-422f0a85f23b37087f1c43d9708e7863455651076e2980dd2285c921a5657168R57

packages/govern-create/utils/network.ts
kanaagawa: This is the default used on the hardhat community 🙈

nivida (approved):

nivida:
Approved because the code quality can be improved in a later state of the project.

PR #256, opened 2020-12-08
Create: add deployments metadata

0xGabi:


nivida:
Do we really need those ABI's in the repo? especially for the localhost?

0xGabi:
These are the metadata generated by the `hardhay-deploy` plugin. I think they are valuable as they have the latests ABIs deployed to the different networks. But you are right we can probably skip localhost 👍

izqui:
@nivida @0xGabi is this still needed? Let's close or merge

novaknole:
Closing this, as it doesn't matter if the repo stores deployment files or not. Each time deploy script is run, it uses `deterministicDeployment` which means even if the deployment json files are not stored, it still creates the same ones.

PR #261, opened 2020-12-10
Console: remove console warnings

bpierre:
@Evalir I tried to separate the changes in small commits, so they can be reviewed one by one. Let me know what you think!


packages/govern-console/src/lib/web3-utils.js
```diff
@@ -95,59 +120,37 @@ export function shortenAddress(address, charsLength = 4) {
...
 }
 
-export function hexToAscii(hexx) {
-  const hex = hexx.toString()
-  let str = ''
-  for (let i = 0; i < hex.length && hex.substr(i, 2) !== '00'; i += 2) {
-    str += String.fromCharCode(parseInt(hex.substr(i, 2), 16))
-  }
-  return str
-}
```
bpierre, line 135: I couldn’t find any use of this function, also I am not sure what `hexx` is supposed to be, other than a `string`? :thinking: 

Let me know if it should stay!

packages/govern-console/src/lib/web3-utils.js
```diff
@@ -9,8 +10,10 @@ export const ETH_EMPTY_HEX = '0x'
 const ETH_ADDRESS_SPLIT_REGEX = /(0x[a-fA-F0-9]{40}(?:\b|\.|,|\?|!|;))/g
 const ETH_ADDRESS_TEST_REGEX = /(0x[a-fA-F0-9]{40}(?:\b|\.|,|\?|!|;))/g
 
-export function bigNum(value) {
-  return new EthersUtils.BigNumber(value)
+export function bigNum(
+  value: string | number | bigint | EthersBigNumber,
```
bpierre: Not sure we need all of this, but I guess it doesn’t hurt.

packages/govern-console/src/environment.ts
```diff
@@ -1,46 +1,57 @@
...
 }
 
-export default function env(name: string) {
-  // @ts-ignore
-  const envVar = ENV_VARS[name]
-  return typeof envVar === 'function' ? envVar() : null
+export default function env<T extends keyof EnvVarHelpers>(
+  name: T,
+): ReturnType<EnvVarHelpers[T]> {
+  return ENV_VARS[name]() as ReturnType<EnvVarHelpers[T]>
```
bpierre, line 56: A headache happened here :brain: :zap:

packages/govern-console/src/Providers/Permissions.tsx
```diff
@@ -41,40 +50,58 @@ type Role = {
...
+        return permissions
+      }
+
+      const roleName = KNOWN_QUEUE_ROLES.get(role.selector)
+      if (roleName && isPermissionsKey(roleName)) {
+        permissions[roleName] = true
+      }
+      return permissions
+    },
+    { ...permissions },
```
bpierre, line 79: A subtle change compared to previously is that we are now copying `permissions`, rather than modifying it. I think this is why [calling `populatePermissions()` in a render](https://github.com/aragon/govern/pull/261/files#diff-690b8b9e540016c91f87342dc92c7c89b5e33076aaf5c380eecc213b03cf16d1L26) could be done without infinite loop.

To prevent another infinite loop to happen, `permissions` is now handled by `useReducer()`.

packages/govern-console/src/Providers/Wallet.tsx
```diff
@@ -4,30 +4,29 @@ import { UseWalletProvider, useWallet, Wallet } from 'use-wallet'
...
 
-type WalletAugmentedContextProps = {
+export type WalletAugmentedData = {
   wallet: Wallet<any>
   ethers: any
-} | null
+}
 
-const WalletAugmentedContext = createContext<WalletAugmentedContextProps>(null)
+const WalletAugmentedContext = createContext<WalletAugmentedData | null>(null)
```
bpierre, line 28: For `WalletAugmentedData` and similar, I made it so that `null` is part of the context, but not the data type itself. This is so that the associated hooks (e.g. `useWalletAugmented()` here) can return the data without `null`, which makes it easier for consumers (since they don’t have to test for `null`).

packages/govern-console/src/components/Header/Header.tsx
```diff
@@ -25,12 +25,12 @@ function Header() {
...
   const handleChangeChain = useCallback(
     e => {
-      setChainId(e.target.value)
+      updateChainId(Number(e.target.value))
       // When we change the chain ID, the DAO might not exist,
       // so we must revert back to the DAO selection screen.
       history.push(`/${getNetworkName(e.target.value)}`)
     },
-    [history, setChainId],
+    [history, updateChainId],
```
bpierre, line 33: I renamed `setChainId` to `updateChainId` to make it more clear that it is not the function from `useState()` but a custom one.

packages/govern-console/package.json
```diff
@@ -16,7 +16,7 @@
     "react-dom": "16.13.1",
     "react-query": "^2.26.2",
     "react-router-dom": "^5.2.0",
-    "react-scripts": "3.4.4",
+    "react-scripts": "4.0.1",
```
Evalir, line 19: Awesome! I also tried upgrading to React 17, but ran into a [bunch of errors](https://github.com/aragon/govern/pull/233/commits/184291f700647b7009e9a9c2626d1d937ef3b1ee). We should try again soon 😄  (but in a separate PR)

packages/govern-console/src/components/Header/Header.tsx
```diff
@@ -121,7 +121,7 @@ function Header(): JSX.Element {
           `}
         >
           {wallet.status === 'connected'
-            ? shortenAddress(wallet.account!)
+            ? shortenAddress(wallet?.account ?? '')
```
Evalir, line 124: Hmm, why is this better? Is it down to style? Or are we just being extra safe here?

packages/govern-console/src/environment.ts
Evalir: Oh my 😅 I gotta read more about how this works!

packages/govern-console/src/lib/web3-utils.js
Evalir: Hmm, I was wondering if we can even just switch to using native BigInts? I haven't tested querying smart contracts with them though.

packages/govern-console/src/lib/web3-utils.js
Evalir: Nope! It's fine if it goes—this file was copied over from the ANJ site I believe, so it'll probably have some things we don't need. :)

packages/govern-console/src/components/Header/Header.tsx
Evalir: Looks good!

packages/govern-console/src/Providers/Wallet.tsx
Evalir: Oh I see why now! Much more ergonomic then. :)

packages/govern-console/src/components/ViewDao/ViewDao.tsx
```diff
@@ -23,7 +23,6 @@ export default function ViewDao({ dao }: ViewDaoProps): JSX.Element {
   const history = useHistory()
   const { ethers } = useWalletAugmented()
   const { permissions, populatePermissions } = usePermissions()
-  populatePermissions(dao.queue.roles)
```
Evalir, line 26: Glad you could remove this... This didn't feel good at all 😅

packages/govern-console/src/Providers/Permissions.tsx
Evalir: oh nice! Feels much cleaner.

Evalir (approved):
This is looking 🔥 just left a few questions mostly. We can of course discuss more deeply when we pair soon! :) (Feel free to merge)

packages/govern-console/src/lib/web3-utils.js
Evalir: And actually, now that we have typings, we could probably move this file and `web3-contracts` to use the `.ts` right?

packages/govern-console/src/components/Header/Header.tsx
bpierre: The `!` tells TypeScript to ignore the fact that `wallet` and `account` can be `null`, so this is something I generally consider as unsafe in TS. Using `?.` and `?? ''` here address this possibility instead, ensuring a string is always returned :-)

packages/govern-console/src/lib/web3-utils.js
bpierre: Oh I completely missed the fact that it wasn’t a `.ts` file! Will do :+1:

bpierre: Done in https://github.com/aragon/govern/pull/261/commits/474834e3b35bb75b3567b2fa45b7723adff36cc1 − that required a few new changes :-)

packages/govern-console/src/lib/web3-contracts.ts
```diff
@@ -1,52 +1,52 @@
...
+    ? new providers.Web3Provider(wallet.ethereum)
+    : getDefaultProvider(getNetworkEthersName(chainId))
 
-  if (!address || !EthersUtils.isAddress(address) || !ethersProvider || !abi) {
-    return null
-  }
+  const ethersSignerProvider =
+    signer &&
+    wallet.ethereum &&
+    ethersProvider instanceof providers.Web3Provider
```
bpierre, line 26: This is because `getDefaultProvider()` might return `BaseProvider` rather than `Web3Provider`, and it doesn’t have `.getSigner()`.

packages/govern-console/src/lib/web3-contracts.ts
```diff
@@ -58,32 +58,31 @@ export function useContractReadOnly(
...
 
-  const cancelBalanceUpdate = useRef(null)
+  const cancelBalanceUpdate = useRef<(() => void) | null>(null)
 
   const updateBalance = useCallback(() => {
     let cancelled = false
 
     if (cancelBalanceUpdate.current) {
-      cancelBalanceUpdate.current()
+      cancelBalanceUpdate.current?.()
```
bpierre: I’m not sure it was happening, but before this change it was always calling `cancelBalanceUpdate()` despite the possibility of it being `null`.

packages/govern-console/src/lib/web3-utils.js
```diff
@@ -1,166 +0,0 @@
...
-    '...' +
-    address.slice(-charsLength)
-  )
-}
-
-export function sanitizeNetworkType(networkType: string): string {
-  if (networkType === 'private') return 'localhost'
-  if (networkType === 'main') return 'mainnet'
-  return networkType
-}
```
bpierre: I removed this

packages/govern-console/src/lib/web3-utils.js
```diff
@@ -1,166 +0,0 @@
...
-
-export function isLocalOrUnknownNetwork(chainId = env('CHAIN_ID')): boolean {
-  return getNetworkType(chainId) === DEFAULT_LOCAL_CHAIN
-}
```
bpierre: Also this got removed

packages/govern-console/src/lib/web3-utils.js
```diff
@@ -1,166 +0,0 @@
...
-
-export function getNetworkType(chainId = env('CHAIN_ID')): string {
-  chainId = String(chainId)
-
-  if (chainId === '1') return 'main'
-  if (chainId === '3') return 'ropsten'
-  if (chainId === '4') return 'rinkeby'
-
-  return DEFAULT_LOCAL_CHAIN
-}
```
bpierre: This got removed in the .ts version

packages/govern-console/src/lib/web3-contracts.ts
Evalir: Nice catch!

packages/govern-console/src/lib/web3-contracts.ts
Evalir: Hmm and can it be null? As we're already checking if the ref is not `falsy` on the `if` statement.

bpierre: True haha, sorry I was having my type-only glasses :eyeglasses:. I guess we could remove the condition now :smile:

Evalir: Haha yes!

(⌐■_■)
( •_•)>⌐■-■
(•_•)

packages/govern-console/src/lib/web3-utils.js
Evalir: 🧹 🧹 🧹

packages/govern-console/src/lib/web3-contracts.ts
bpierre: Done in https://github.com/aragon/govern/pull/261/commits/83b0b02f27887c0ee36f7332b27f30a1cef9fac2

(I also moved the `cancelBalanceUpdate.current = null`)

PR #262, opened 2020-12-10
Console: fix network selection + other fixes

bpierre:


packages/govern-console/src/components/Header/Header.tsx
```diff
@@ -24,11 +24,14 @@ function Header(): JSX.Element {
   }, [chainId, wallet])
 
   const handleChangeChain = useCallback(
-    e => {
-      updateChainId(Number(e.target.value))
+    (event: { target: { value: string } }) => {
```
bpierre, line 27: For some reason TS wasn’t complaining about not knowing the type of `e.target.value`, and passing it to `getNetworkName()`.

Evalir (approved):
🏅 🏅 🏅

PR #263, opened 2020-12-11
govern-core: check container execution time is valid on GovernQueue#resolve()

sohkai:
Nice catch, @nivida!

### DIFF
--- a/packages/govern-core/contracts/pipelines/GovernQueue.sol
+++ b/packages/govern-core/contracts/pipelines/GovernQueue.sol
@@ -134,9 +134,6 @@ contract GovernQueue is IERC3000, IArbitrable, AdaptiveERC165, ACL {
         auth(this.execute.selector) // in most instances this will be open for any addr, but leaving configurable for flexibility
         returns (bytes32 failureMap, bytes[] memory)
     {
-        // ensure enough time has passed
-        require(block.timestamp >= _container.payload.executionTime, "queue: wait more");
-
         bytes32 containerHash = _container.hash();
         queue[containerHash].checkAndSetState(
             GovernQueueStateLib.State.Scheduled, // note that we will revert here if the container wasn't previously scheduled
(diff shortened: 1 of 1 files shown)


nivida (approved):

izqui:
I might be missing something, but this was intentional.

After the resolver has made a final ruling, there's no reason to keep waiting since the container cannot be challenged more than once. Due to this, we assume that it can be immediately executed.

nivida:
@izqui Our point was that the configured execution delay does no longer work as expected without checking it in the resolve method of the ``GovernQueue``. From my perspective and also from Marc should the voting get re-scheduled instead of executing it. This to keep the delay some users probably will count on (e.g.: "let me check that later"). And that someone can challenge for example another part of the voting. One will probably challenge the proof and another one the included actions idk.

I'm currently not 100% sure about the internals of the court used to resolve a dispute but is there 100% proof that a resolved dispute always will be correct? :-)

izqui:
Thanks for the context, I still think that this delay is not needed since we should only allow one dispute, otherwise you can DOS this with ongoing challenges.

If an action is approved then it should be executable immediately, even if the evidence provided for the challenge is for one small part, guardians should always check the validity of the entire payload.

nivida:
Sounds reasonable to me! Thanks for your input, Jorge! I will close this PR.

marcvi1a:
@izqui I did not understand your response. Let me write the use case again: 

Disputable delay is 7 days, so users count on having a full week to review any action that's going to move on-cain.

Vote on a new action actually passes and goes into the queue. It's day 1 and someone challenges the action. The dispute is resolved in 24 hours, the claim is bad and the action must go on. So we're now in day 2 and the action has been resolved, should it be already implemented on-chain? That would "speed up" the process and cut the disputable delay from 7 days down to 2. Perhaps some users are no in favor of that action and would like to sell their tokens, for which they're counting on having 7 days as usual.

Therefore I think the original 7 days should be respected. I would propose for that action to move back to the queue in order to be implemented on-chain in 5 days time. It could be marked as already revised so no-one else can open a dispute.

izqui:
I understand your point, but once it is challenged, guardians shouldn't just look at what the challenger claims but whether the entire action is valid.

Regardless of this wait, actions can only be challenged once in the code so during this additional waiting period it couldn't be re-challenged as it currently stands.

marcvi1a:
I agree actions can only be challenged once. But to challenge an action should never be a mechanism to speed up implementation of that action. That's why the action needs to go back to the queue, and get implemented only after the delay period has concluded.

PR #272, opened 2020-12-14
Core(Queue): add relative gas limit checks to prevent DOS attacks

### LINKED ISSUE #265: CD 4.1: Gas DoS in GovernQueue.schedule allows anyone to create unchallengable Containers

izqui:
Closes #265

### DIFF
--- a/packages/govern-core/contracts/pipelines/GovernQueue.sol
+++ b/packages/govern-core/contracts/pipelines/GovernQueue.sol
@@ -63,6 +63,14 @@ contract GovernQueue is IERC3000, IArbitrable, AdaptiveERC165, ACL {
     mapping (bytes32 => address) public challengerCache; // container hash -> challenger addr (used after challenging and before dispute resolution)
     mapping (bytes32 => mapping (IArbitrator => uint256)) public disputeItemCache; // container hash -> arbitrator addr -> dispute id (used between dispute creation and ruling)
 
+    modifier maxRelativeGas(uint256 allowedGasPct) {
+        uint256 initialGas = gasleft();
+        _;
+        uint256 gasSpent = initialGas - gasleft();
+        uint256 maxGas = block.gaslimit() * allowedGasPct / 100;
+        require(gasSpent <= maxGas, "queue: gas");
+    }
+
     /**
      * @param _aclRoot account that will be given root permissions on ACL (commonly given to factory)
      * @param _initialConfig initial configuration parameters
@@ -88,6 +96,7 @@ contract GovernQueue is IERC3000, IArbitrable, AdaptiveERC165, ACL {
         public
         override
         auth(this.schedule.selector) // note that all functions in this contract are ACL protected (commonly some of them will be open for any addr to perform)
+        maxRelativeGas(33) // DOS protection: To ensure that the container can always be challenged
         returns (bytes32 containerHash)
     {
         // prevent griefing by front-running (the same container is sent by two different people and one must be challenged)
@@ -258,6 +267,7 @@ contract GovernQueue is IERC3000, IArbitrable, AdaptiveERC165, ACL {
         public
         override
         auth(this.configure.selector)
+        maxRelativeGas(20) // DOS protection: to ensure that the config doesn't take up a lot of space in containers
         returns (bytes32)
     {
         return _setConfig(_config);


nivida:
This fix does look good to me.

But couldn't we also refactor the ``challenge`` method with for example just using the hash as dispute metadata and leaving the evidence period open?

This means the ``challenge`` method would just "register" a dispute and all the evidence does have to get added after. A possible method signature with such a solution would be:

```function challenge(bytes32 containerHash, ERC3000Data.Config memory config, bytes memory _reason)```

A clear disadvantage I see from such a solution is the decreasing DX/UX. I just thought to bring this possible solution up as well.

**Comment by izqui** [comment]
@nivida can you do a prototype of what you propose to see how it looks? I'd rather keep it simple tbh

**Comment by nivida** [comment]
@izqui I totally agree to keep it simple. I will open a small PR asap just to "visualize" another possible solution. But don't think mine will be the winning one :-)

**Review by sohkai — COMMENTED** [review]
It may be worth considering allowing the relative percentages to be dynamically configurable:

- Gas costs can change in the future
- `challenge()`'s gas cost multipler is strictly dependent on the arbitrator being used

**Comment by nivida** [comment]
> It may be worth considering allowing the relative percentages to be dynamically configurable:
> 
> * Gas costs can change in the future
> * `challenge()`'s gas cost multipler is strictly dependent on the arbitrator being used

Totally agree. This was once what I was mention in a private chat with Jorge.

**Comment by novaknole** [comment]
Hi all, 

Just wanted to show one example where it could go wrong.

I will follow the same gas costs with the 33% percentage as the `maxRelativeGas` for the `schedule` function as shown in this pull request.

Schedule   - `Cost >= 24.3(P) + [(1/512)*(P/32)^2]`
Challenge  - `Cost >= 33.2(P) + [(3/256)*(P/32)^2]`

the `maxRelativeGas ` modifier checks the gas spent only for the function body which means it doesn't include `intrinsicGas` + `calldatacopy` that happens on the entry of the function since modifier starts to get called after these 2 and after some opcodes(not important here)..

So, we can separate `intrinsicGas` + `calldatacopy` on entry point costs from the actual body costs. 

Schedule   - `Cost >= 8.09P + | 16.18P + (P/32)^2 / 512`
Challenge  - `Cost >= 16.93P + (5/512) *(P/32)^2   + | 16.18P + (P/32)^2 / 512`

The `intrinsicGas` + `calldatacopy` on entry point costs are after `|` mark. They are the same for sure..

If `P` is 400000, we get

Schedule   - `Cost >= 3236000   + | + 6777175 = 10013175`
Challenge  - `Cost >= 8297878   +  | + 6777175 = 15075053`

If the gas limit is 15000000, the `33%` of it is 4950000 which means our `schedule` function will be succeeded for `P` = 400000, because the function body needed `3236000`, but `Challenge` now needs 15075053 which is more than gas limit.

I might be missing something for sure,  so, would love your opinions on this too.

================================

So, if we still choose the `gas modifier` checks,  it's notable that instead of `P`, we can presume that it's `2P`, due to the fact that in the `ERC3000Data` type, we have 2 `bytes` parameter.   So, after playing with it , one can determine that the percentage we can choose looks like this:

```
K = 1620 / (66 + 0.0000457P) 
```

So, if the `P` is 400,000 , Then, `K` should be less than 19.3% .  Also, all these costs don't take into account how much gas `createDispute` call will take on its own since the arbitrator could be other contract than `DisputeManager`, but in this examples and from the audit, it's presumed to be constant. 

If the gas costs change, we still have to find/calculate the good percentage again and again and update the value which seems to be a little bit complex.

I still would stick with one of these solutions, depending on the fact that we must pass the whole container to the artibtrator and this is a MUST.

* predefined length (really small) for those two `bytes` parameter. Even if `P` is 10,000 bytes maximum length,  everything still should be safe, but we would need `require` to check it, but sure, dynamic gas modifier check solution is more dynamic.
* maybe, on those `bytes` parameters, we put the hashes instead of the actual proofs ? proofs could be stored on the IPFS. 



Question 1. what's the actual reason for using `memory` type in the function's parameters ? since extra copy is inevitable in this case. The function could be `external` .

izqui:
I am no longer that convinced that relative gas checks are a good idea, it feels like a bad hack.

We should explore alternatives. A configurable max size for the arbitrary size data pieces might be a better idea.

PR #276, opened 2020-12-16
CI/CD: Revert of accidentally merged PRs or updating of tests

nivida:
This PR does highlight the problems we have currently with our CI/CD. In the case of the broken subgraph did I just out-comment the registration of the event handler. To actually fix the subgraph is another PR required with a well-defined scope.

The Problems:

- PR #259 got merged without discussing the proposed changes or at least adjusting the tests 
- The ``IArbitrable`` interface got adjusted to reflect Aragon Protocol v2. This means the ``Arbitrable`` does no longer have the ``EvidenceSubmitted`` event. This event is now within the ``DisputeManager``. 
    - @0xGabi Means we would have to add the ``DisputeManager`` to the subgraph of Govern and listen there for the events. 

Would be great if we could discuss here the changes which did get merged without a real review or without updating the Govern Framework on the points the changes did have an impact on.


nivida:
I'm closing this PR because @novaknole has opened the required PR's as described in the description of this PR.

PR #277, opened 2021-02-04
remove evidenceSubmitted event

novaknole:
This removes `evidenceSubmitted` event since it's been removed from the `IArbitrable` and is now moved to `DisputeManager`. 

Also, there's `handleRuled` mapping which was doing the following:

```
let containerHash = governQueue.disputeItemCache(
   event.params.arbitrator,
   event.params.disputeId
)
```

`disputeItemCache` has been changed and now expects arguments as `containerHash` and `disputeId`.. The `Ruled` event, though, doesn't emit `containerHash`, so it's impossible to grab the container.  

So, this also removes `handleRuled` mapping and event. What `handleRuled` intended to do is now done by `handleResolved`.

### DIFF
--- a/packages/govern-subgraph/src/GovernQueue.ts
+++ b/packages/govern-subgraph/src/GovernQueue.ts
@@ -145,26 +145,6 @@ export function handleConfigured(event: ConfiguredEvent): void {
   queue.save()
 }
 
-// IArbitrable Events
-
-
-
-export function handleRuled(event: RuledEvent): void {
-  // let governQueue = GovernQueueContract.bind(event.address)
-  // let containerHash = governQueue.disputeItemCache(
-  //   event.params.arbitrator,
-  //   event.params.disputeId
-  // )
-  // let container = loadOrCreateContainer(containerHash)
-
-  // container.state =
-  //   event.params.ruling === ALLOW_RULING ? APPROVED_STATUS : REJECTED_STATUS
-
-  // handleContainerEventRule(container, event)
-
-  // container.save()
-}
-
 // MiniACL Events
 
 export function handleFrozen(event: FrozenEvent): void {
--- a/packages/govern-subgraph/src/GovernQueue.ts
+++ b/packages/govern-subgraph/src/GovernQueue.ts
@@ -10,7 +10,6 @@ import {
   Scheduled as ScheduledEvent,
   Vetoed as VetoedEvent,
   Ruled as RuledEvent,
-  EvidenceSubmitted as EvidenceSubmittedEvent,
   GovernQueue as GovernQueueContract
 } from '../generated/templates/GovernQueue/GovernQueue'
 import {
@@ -38,7 +37,6 @@ import {
   handleContainerEventResolve,
   handleContainerEventRule,
   handleContainerEventSchedule,
-  handleContainerEventSubmitEvidence,
   handleContainerEventVeto
 } from './utils/events'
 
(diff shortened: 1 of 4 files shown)


packages/govern-subgraph/src/GovernQueue.ts
```diff
@@ -95,7 +93,7 @@ export function handleChallenged(event: ChallengedEvent): void {
 export function handleResolved(event: ResolvedEvent): void {
   let container = loadOrCreateContainer(event.params.containerHash)
 
-  container.state = event.params.approved ? EXECUTED_STATUS : CANCELLED_STATUS
+  container.state = event.params.approved ? APPROVED_STATUS : REJECTED_STATUS
```
nivida, line 96: If I'm right should it be ``EXECUTED_STATUS`` and ``CANCELLED_STATUS`` or can you elaborate closer why not?

packages/govern-subgraph/src/GovernQueue.ts
```diff
@@ -147,35 +145,6 @@ export function handleConfigured(event: ConfiguredEvent): void {
...
-  let containerHash = governQueue.disputeItemCache(
-    event.params.arbitrator,
-    event.params.disputeId
-  )
-  let container = loadOrCreateContainer(containerHash)
-
-  handleContainerEventSubmitEvidence(container, event)
-}
-
-export function handleRuled(event: RuledEvent): void {
```
nivida, line 163: The ``Ruled`` event is still existing on the ``GovernQueue`` [here](https://github.com/aragon/govern/blob/master/packages/govern-core/contracts/pipelines/GovernQueue.sol#L213). 

Was there another reason why this got removed?

nivida (changes requested):
Added two comments because I think too much got removed. Tuned to hear your opinion about @novaknole

packages/govern-subgraph/src/GovernQueue.ts
novaknole: This is where the `resolved` event gets thrown. https://github.com/aragon/govern/blob/2557caba58170c4625e07d19f1f27202dc19d139/packages/govern-core/contracts/pipelines/GovernQueue.sol#L212  

As we can see, the third parameter is `ArbitratorApproved`. So, it's all about approving or rejecting and not executing/cancelling.

packages/govern-subgraph/src/GovernQueue.ts
novaknole: `HandleRuled` event seems to be changed.  In the subgraph mapping, this is what we had:

```
let containerHash = governQueue.disputeItemCache(
    event.params.arbitrator,
    event.params.disputeId
  )
```

Since `disputeItemCache` was changed too, meaning that the first key we should pass is `containerHash`, it's impossible to get containerHash from the above code. Which means `Ruled` event doesn't help at all.

packages/govern-subgraph/src/GovernQueue.ts
nivida: Thanks for the discussion on Discord. Anythings looks good 👍

packages/govern-subgraph/src/GovernQueue.ts
nivida: Thanks for the discussion on Discord. Anything looks good 👍

nivida:
Merged with approval from me by @novaknole.

PR #282, opened 2021-02-10
test merkledistributor

### LINKED ISSUE #163: Token: test MerkleDistributor.sol
Bring tests from https://github.com/Uniswap/merkle-distributor/

novaknole:
This brings the tests for `MerkleDistributor.sol` and solves #163 

A couple of notes:

*  In the `MerkleDistributor.sol`, there was a line which seemed wrong.

```
return !MerkleProof.verify(_proof, merkleRoot, node);
```

It should have been the below without `!` operator.

```
 return MerkleProof.verify(_proof, merkleRoot, node);
```

* I went ahead and removed the tests which tested `gasUsed` . They don't do anything really good.
* Seems like there's no tests for `GovernMinter`.  Is there a requirement for this or no ?
* 2 helper files that are used by the tests are `ts` files. Hope that's not a problem.

### DIFF
--- a/packages/govern-token/contracts/MerkleDistributor.sol
+++ b/packages/govern-token/contracts/MerkleDistributor.sol
@@ -12,6 +12,7 @@ import "@aragon/govern-contract-utils/contracts/erc20/SafeERC20.sol";
 import "@aragon/govern-contract-utils/contracts/initializable/Initializable.sol";
 
 contract MerkleDistributor is Initializable {
+    
     using SafeERC20 for ERC20;
 
     ERC20 public token;
@@ -48,7 +49,7 @@ contract MerkleDistributor is Initializable {
 
     function _verifyBalanceOnTree(uint256 _index, address _to, uint256 _amount, bytes32[] memory _proof) internal view returns (bool) {
         bytes32 node = keccak256(abi.encodePacked(_index, _to, _amount));
-        return !MerkleProof.verify(_proof, merkleRoot, node);
+        return MerkleProof.verify(_proof, merkleRoot, node);
     }
 
     function isClaimed(uint256 _index) public view returns (bool) {
--- a/packages/govern-token/contracts/test/TestERC20.sol
+++ b/packages/govern-token/contracts/test/TestERC20.sol
@@ -19,3 +19,4 @@ contract TestERC20 is ERC20 {
         }
     }
 }
+
--- a/packages/govern-token/contracts/test/TestERC20.sol
+++ b/packages/govern-token/contracts/test/TestERC20.sol
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.6.8;
+
+import '@openzeppelin/contracts/token/ERC20/ERC20.sol';
+
+contract TestERC20 is ERC20 {
+    constructor (string memory name_, string memory symbol_, uint amountToMint) ERC20(name_, symbol_) public {
+        setBalance(msg.sender, amountToMint);
+    }
+
+    // sets the balance of the address
+    // this mints/burns the amount depending on the current balance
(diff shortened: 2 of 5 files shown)


packages/govern-token/contracts/MerkleDistributor.sol
```diff
@@ -48,7 +49,7 @@ contract MerkleDistributor is Initializable {
 
     function _verifyBalanceOnTree(uint256 _index, address _to, uint256 _amount, bytes32[] memory _proof) internal view returns (bool) {
         bytes32 node = keccak256(abi.encodePacked(_index, _to, _amount));
-        return !MerkleProof.verify(_proof, merkleRoot, node);
+        return MerkleProof.verify(_proof, merkleRoot, node);
```
nivida, line 52: Can you elaborate closer on why inverting isn't required?

packages/govern-token/contracts/test/TestERC20.sol
```diff
@@ -0,0 +1,21 @@
...
+    // this mints/burns the amount depending on the current balance
+    function setBalance(address to, uint amount) public {
+        uint old = balanceOf(to);
+        if (old < amount) {
+            _mint(to, amount - old);
+        } else if (old > amount) {
+            _burn(to, old - amount);
+        }
+    }
+}
```
nivida: EOF error

packages/govern-token/test/helpers/merkle-tree/balance-tree.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  public getHexRoot(): string {
+    return this.tree.getHexRoot()
+  }
+
+  // returns the hex bytes32 values of the proof
+  public getProof(index: number | BigNumber, account: string, amount: BigNumber): string[] {
+    return this.tree.getHexProof(BalanceTree.toNode(index, account, amount))
+  }
+}
```
nivida: EOF error

packages/govern-token/test/helpers/merkle-tree/merkle-tree.ts
```diff
@@ -0,0 +1,124 @@
...
+      throw new Error('Array is not an array of buffers')
+    }
+
+    return arr.map((el) => '0x' + el.toString('hex'))
+  }
+
+  private static sortAndConcat(...args: Buffer[]): Buffer {
+    return Buffer.concat([...args].sort(Buffer.compare))
+  }
+}
```
nivida: EOF error

nivida (commented):
> I went ahead and removed the tests which tested gasUsed . They don't do anything really good.

I think testing of ``gasUsed`` was here to have a kind of monitoring. But I will verify this assumption now.

> Seems like there's no tests for GovernMinter. Is there a requirement for this or no ?

Feel free to create an issue about the missing test cases.

> 2 helper files that are used by the tests are ts files. Hope that's not a problem.

Should be good.

I have added a question to understand better why exactly the inversion of the return value in ``_verifyBalanceOnTree`` got removed to later on being able to do a deeper review of this PR.

packages/govern-token/contracts/MerkleDistributor.sol
novaknole: Let's say that `verify` returns true. If so, `_verifyBalanceOnTree` would return false. Then, if we look at this check:

https://github.com/aragon/govern/blob/55ccf385b28d5d05443d5170b0e463ccc145ffd7/packages/govern-token/contracts/MerkleDistributor.sol#L37 we can see that in the `require`, it would get `false` and the whole transaction would fail even though it should have been succeeded.

nivida: Thanks for the closer explanation! :)

nivida (approved):
Approved to merge if the EOF lines do get added.

PR #284, opened 2021-02-24
fix token factory with zero address

### LINKED ISSUE #283: creating DAO with 0x..00 address and without proxies fail
by calling `newGovernWithoutConfig`, anyone can create a new DAO. They can pass `address(0)` if they want the Govern to create tokens for them, or they can already pass the created token that they are already using. This problem is about when they pass `address(0)` and when they don't want to use proxies.

For example:

```
newGovernWithoutConfig("DAO-7", "0x0000000000000000000000000000000000000000", "TokenName", "TKN", false)
```

The problem seems to be : **execution reverted: acl: auth** because of this https://github.com/aragon/govern/blob/5c0293fda66c188b971f96de0666b6309e379c78/packages/govern-token/contracts/GovernTokenFactory.sol#L58

As shown, `GovernTokenFactory` calls mint on the minter, but `GovernTokenFactory` doesn't have the rights. The only party that has this right is `GovernBaseFactory`.

---

Solution seems to be:

https://github.com/aragon/govern/blob/5c0293fda66c188b971f96de0666b6309e379c78/packages/govern-token/contracts/GovernTokenFactory.sol#L40

the first argument should be address(this) instead of _initialMinter

This problem doesn't show when we use proxies, because for proxies, this is exactly what happens. https://github.com/aragon/govern/blob/5c0293fda66c188b971f96de0666b6309e379c78/packages/govern-token/contracts/GovernTokenFactory.sol#L52

novaknole:
This presumably fixes the issue that I found and created alongside this pull request. #283 

Could be my mistake though..


izqui (approved):

packages/govern-token/contracts/GovernTokenFactory.sol
```diff
@@ -37,7 +37,7 @@ contract GovernTokenFactory {
         GovernMinter minter
     ) {
         if (!_useProxies) {
-            (token, minter) = _deployContracts(_initialMinter, _tokenName, _tokenSymbol, _tokenDecimals);
+            (token, minter) = _deployContracts(address(this), _tokenName, _tokenSymbol, _tokenDecimals);
```
nivida: Thanks for sorting this out! :) 

We can remove the ``_initialMinter`` argument from the ``_deployContracts`` method at all and just use ``address(this)`` directly in the ``_deployContracts`` method. This will for sure save a bit of gas.

Another problem we have here is that in the end the ``GovernBaseFactory`` is the defined minter. See lines 65-66 here and line 50 in ``GovernBaseFactory``. I think the goal would have been to grant the permissions to ``_mintAddr`` and not ``_initialMinter``. @novaknole Thanks for asking back related to the second problem we have here.

nivida (changes requested):

nivida (approved):

PR #293, opened 2021-03-08
Update buidler

novaknole:
https://linear.app/aragon/issue/ARA2-61/govern-token-update-buidler-to-hardhat


packages/govern-token/deploy/factory.ts
```diff
@@ -0,0 +1,23 @@
+const { print } = require('../lib/utils')
+
```
nivida: So many breaks aren't required.

packages/govern-token/hardhat.config.ts
```diff
@@ -0,0 +1,60 @@
+import { config as dotenvConfig } from 'dotenv'
+import { resolve } from 'path'
+dotenvConfig({ path: resolve(__dirname, './.env') })
```
nivida: Could you not also have it after the imports?

packages/govern-token/lib/utils.js
```diff
@@ -5,4 +5,4 @@ const format = ({ address }, name) =>
 
 const print = (contract, name) => console.log(format(contract, name))
 
-module.exports = { format, print }
+module.exports = { format, print }
```
nivida: EOF

nivida (changes requested):

packages/govern-token/deploy/factory.ts
novaknole: Missed this. Thanks ^__^

PR #296, opened 2021-03-10
addressutils tests + fix isContract

### LINKED ISSUE #114: Utils: Write test for AddressUtils library

novaknole:
It seems like the code was using `not` assembly instruction for the 2 things.

* ```not(iszero(mload(ptr)))```
* ```not(iszero(extcodesize(addr)))```

The idea is that both of the equations above get assigned to `bool` variables. `not` is not the NOT operator. It's a bitwise `NOT` operator of `x`, which means `not(1)` won't be 0, which means that the above 2 are not correct.

We change them to be:

* ```iszero(iszero(mload(ptr)))```
* ```iszero(iszero(extcodesize(addr)))```

@nivida  Thanks to Sam for discussing this with me too.

@izqui your thoughts ? 

Fixes #114

### DIFF
--- a/packages/govern-contract-utils/contracts/address-utils/AddressUtils.sol
+++ b/packages/govern-contract-utils/contracts/address-utils/AddressUtils.sol
@@ -5,6 +5,7 @@
 pragma solidity ^0.6.8;
 
 library AddressUtils {
+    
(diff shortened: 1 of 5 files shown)


nivida (approved):
LGTM! :)

PR #304, opened 2021-03-17
chore(govern-create): add optional param to `deploy-govern`

mariocao:
This PR adds an optional parameter to the `deploy-govern` hardhat task in order to allow specifying not only the factory but also the registry.

Btw, I hope you don't mind I updated the Rinkeby addresses (those that are being tracked by the [Subgraph](https://thegraph.com/explorer/subgraph/aragon/aragon-govern-rinkeby) instance).

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,8 @@ To use Govern, feel free to deploy your own registries and factories, but using 
 
 #### Rinkeby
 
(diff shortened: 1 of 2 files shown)


novaknole (approved):
Looks good to me. 

NOTE: The rinkeby addresses will anyways change in the end when we deploy the latest contracts.

PR #306, opened 2021-03-18
Fix/gas dos

novaknole:
This is a different implemenation of what gas limit modifier was doing with `gasleft()` check.

The problem is it requires to change `ERC3000` to add new variable to the config.


novaknole:
Closing this as the new PR will be made to merge into `develop` instead of `master`.

PR #307, opened 2021-03-19
fix null value issue with registryEntry query

yuetloo:
This is to fix the error from RegistryEntry subgraph query error:

```
Null value resolved for non-null field `queue`
```

The way to reproduce this issue is to registry a Dao (call contract GovernRegistry.register()) with an invalid GovernQueue address.

If the name is 'M', then the following subgraph query will give the null error.
```
registryEntry(id: "M") {
      id
      name
    queue {
      id
    }
      
  }
```

Govern does not have the problem because the register() function has a call to _setMetadata() which trigger an event and trigger the subgraph to instantiate the `executor` entity.

@novaknole , I'm not sure if I need to add GovernQueue in the data source entity here after this change:
https://github.com/aragon/govern/blob/master/packages/govern-subgraph/manifest/templates/contracts/GovernRegistry.template.yaml#L13

Doesn't give me any error without adding that to the entities field.

### DIFF
--- a/packages/govern-subgraph/src/GovernQueue.ts
+++ b/packages/govern-subgraph/src/GovernQueue.ts
@@ -1,4 +1,4 @@
-import { Address, Bytes } from '@graphprotocol/graph-ts'
+import { Address, Bytes, BigInt } from '@graphprotocol/graph-ts'
 import {
   Challenged as ChallengedEvent,
   Configured as ConfiguredEvent,
@@ -23,6 +23,7 @@ import {
 import { frozenRoles, roleGranted, roleRevoked } from './lib/MiniACL'
 import { buildId, buildIndexedId } from './utils/ids'
 import {
+  ZERO_ADDRESS,
   APPROVED_STATUS,
   CANCELLED_STATUS,
   CHALLENGED_STATUS,
@@ -39,6 +40,7 @@ import {
   handleContainerEventSchedule,
   handleContainerEventVeto
 } from './utils/events'
+
 
 export function handleScheduled(event: ScheduledEvent): void {
   let queue = loadOrCreateQueue(event.address)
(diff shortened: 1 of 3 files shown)


novaknole:
I think, we are making a mistake here. 

If someone calls `register` on the `GovernRegistry`, it means that they already have a `queue` , `token`, `executor` instances which they pass to `register` function. 

@yuetloo in your case, you're creating a dummy config and assign it, which shouldn't be done like this. We should grab the original queue with `queue.bind(addressOfQueue)` and then get a config from there and create the same config from it...


Creating a dummy config will result in a wrong data in the subgraph, because if we ask subgraph to return the queue and its config, it will return dummy config even though it should be returning the actual config that exists on the queue that user passed to `register` on the `GovernRegistry`.

yuetloo:
@novaknole , the issue was the original `queue` creation event was either not triggered (i.e. someone called `register` with an invalid `queue`) or the creation event was missed because the subgraph was configured to start from a block later than the queue creation.

We had to add a `dummy` queue entity in subgraph when processing the `register` event.

Also, the `queue.bind(addressOfQueue)` only allows you to access the contract state or function, but, only `configHash` was available in the contract as only `configHash` was stored, not the actual config. So, we won't be able to use `bind` to get the actual config content.

novaknole:
After the talk with @yuetloo , We decided to merge this. 

1. if the queue was created with different contract, rather than our `GovernQueue` and if `Configured` event is not thrown, it means that this solution will only set a `dummyConfig` on a queue which is still true hence no configuration event was thrown from the queue.

2. if the queue was created as `GovernQueue` instance, and `register` on `GovernRegistry` was called, it means first `dummyConfig` will be set on a queue, but as soon as the data source template code such as this: `GovernQueueTemplate.create(event.params.queue)` happens, it will automatically overwrite the dummy config and still result in a correct result.

PR #308, opened 2021-03-22
Dao queries for Govern API

yuetloo:
This PR is not ready yet as I still need to update the e2e test to point to the correct subgraph.. I believe the tests used to point to the mainnet subgraph, at least the queyTest passed with the mainnet subgraph.  

mainnet subgraph it used to point to:
https://graph.backend.aragon.org/subgraphs/name/aragon/aragon-govern-mainnet

currently using my own subgraph pending the fixes in #307 .
https://thegraph.com/explorer/subgraph/yuetloo/aragon-govern-v2-rinkeby?version=current

So, still need to adjust the subgraph url and test fixture data.


novaknole:
I will review this once the #307 has been resolved.

yuetloo:
@novaknole , please review this PR. All tests passed now and I set the default configuration to point to the mainnet subgraph just like it was before when it was using the govern-server.

One interesting observation when testing against the mainnet which currently does not have any entry in the GovernRegistry but has 1 entry in the Govern entity.

Because DAOS returns the Governs entity but DAO returns the GovernRegistry entity (because it needs to filter by name and name only exists in GovernRegistry), I noticed that in mainnet, DAOS returns 1 entry even though you'll not be able to search that by name.

yuetloo:
closing this PR to open another one to merge to development branch instead of master branch.

PR #310, opened 2021-03-31
Fix/gas dos

novaknole:


nivida (approved):
LGTM! Thanks for doing those changes. Have you thought about the potential implications the config change can have on console.govern and also on the Snapshot plugin?

PR #311, opened 2021-03-31
Dao queries for Govern API

yuetloo:
Recreating this PR to merge the changes to the development branch.
This PR addresses issues:
1) https://linear.app/aragon/issue/ARA2-30/governjs-%E2%88%92-fetch-data-from-the-subgraph
2) https://linear.app/aragon/issue/ARA2-31/govern-subgraph-%E2%88%92-implement-dao-daos-queries

In addition to the above issues, I also added a workflow to automatically run the e2e tests and unit tests in the govern project.


novaknole:
@yuetloo I add my reviews.

One thing I think we have to worry about is that we are using `rinkeby` network to test the subgraph and using `GIORGI-DAO3`.  If the Rinkeby subgraph gets updated, we also have to update the tests..


Do we have any other way to test it without interfering with `rinkeby` network at all ?

yuetloo:
> @yuetloo I add my reviews.
> 
> One thing I think we have to worry about is that we are using `rinkeby` network to test the subgraph and using `GIORGI-DAO3`. If the Rinkeby subgraph gets updated, we also have to update the tests..
> 
> Do we have any other way to test it without interfering with `rinkeby` network at all ?

That's a good point, I thought about it, that's why I pulled out the fixture data into a json file, thinking to just update the json if there's any changes.  

One option is, we could start a local subgraph/ganache instance to do the e2e tests instead of testing against Rinkeby... just more setup and won't be able to verify some of the addresses and/or hashes I think.. Will need to generate data each time and  register new govern/queue.
If we want to go this route, I suggest create a separate ticket for this.

packages/govern/internal/actions/DaoAction.ts
```diff
@@ -23,6 +28,20 @@ export default class DaoAction extends AbstractAction {
...
+   * Format the result to Doa
+   *
+   * @method format
+   *
+   * @returns {Promise<Dao>}
+   *
+   * @protected
+   */
+  protected format( { registryEntries }: { registryEntries: DaoEntry[] } ): Dao | null {
+
```
nivida: Not required break

packages/govern/internal/actions/DaoAction.ts
```diff
@@ -23,6 +28,20 @@ export default class DaoAction extends AbstractAction {
     super(parameters)
   }
 
+  /**
+   * Format the result to Doa
+   *
+   * @method format
+   *
+   * @returns {Promise<Dao>}
```
nivida: Params are missing in the FuncDoc.

packages/govern/internal/actions/DaoAction.ts
```diff
@@ -1,6 +1,11 @@
 import AbstractAction from './lib/AbstractAction'
 import dao, { Dao } from '../clients/graphql/queries/dao'
 
+interface DaoEntry {
```
nivida: Is it not possible to change the DAO interface imported from ``'../clients/graphql/queries/dao'`` ?

packages/govern/internal/actions/lib/AbstractAction.ts
```diff
@@ -54,6 +54,19 @@ export default abstract class AbstractAction {
     return parameters
   }
 
+  /**
+   * Perform custom data formating, default no formatting
+   *
+   * @method resolve
```
nivida: wrong method name

packages/govern/internal/actions/lib/AbstractAction.ts
```diff
@@ -67,5 +80,6 @@ export default abstract class AbstractAction {
    */
   public execute(): Promise<any> {
```
nivida: Execute can get defined as ``async`` which in the end would result in cleaner code.

packages/govern/tests/e2e/queryTest.e2e.ts
```diff
@@ -1,24 +1,27 @@
...
           id
           address
           metadata
         }
       }
     `)
 
-    expect(response.daos[0].address).toEqual('0x24319b199e9e3867ede90eaf0fad56168c54d077')
+    expect(ethers.utils.isAddress(response.daos[0].address)).toEqual(true)
+    // expect(response.daos[0].address).toEqual('0x24319b199e9e3867ede90eaf0fad56168c54d077')
```
nivida: Can we remove this comment?

.github/workflows/govern-ci-cd.yml
```diff
@@ -0,0 +1,35 @@
+name: Govern API tests
```
nivida, line 1: Can you elaborate closer here or in the PR desc why it was required to add this workflow within this scope?

packages/govern/internal/actions/DaoAction.ts
```diff
@@ -35,6 +54,6 @@ export default class DaoAction extends AbstractAction {
   public async execute(): Promise<Dao> {
```
nivida: We can remove this overwrite at all. This was only required because I was formating the output.

nivida (changes requested):

.github/workflows/govern-ci-cd.yml
yuetloo: the original test workflow was removed from this project somehow.  Since I'm making changes in the project and I want to run the unit/e2e tests, I thought it's the best time to add the workflow as part of this PR.

yuetloo: I'll explain in the PR description as wel.

packages/govern/internal/actions/DaoAction.ts
yuetloo: I'll rename it to Executor and move the interface to ../clients/graphql/queries/dao as it's very confusing registryEntry is used to mean the GovernRegistry entry, the Govern entry, and the GovernQueue entry because of how they are defined in the subqueue.

I also try not to change the DAO interface so that there's minimal change/impact to the DAO interface documented in the guide.  

Also, it helps me do the test by not changing the output interface, DAO.  If I define the query incorrectly, the tests would catch the mistake immediately as the output no longer match expected values.

packages/govern/internal/actions/DaoAction.ts
yuetloo: done

packages/govern/tests/e2e/queryTest.e2e.ts
yuetloo: done

packages/govern/internal/actions/DaoAction.ts
yuetloo: I remove the format() call instead and just use execute() to format the output as the original design, which is better as it's less code..

packages/govern/internal/actions/DaoAction.ts
yuetloo: removed this function and do the formatting in the execute() call instead.

.github/workflows/govern-ci-cd.yml
nivida: Perf. thanks.. and why ever this got removed 🤷

packages/govern/internal/actions/DaoAction.ts
```diff
@@ -23,18 +23,20 @@ export default class DaoAction extends AbstractAction {
...
    *
    * @method execute
    *
-   * @returns {Promise<Dao>}
+   * @returns {Promise<Dao|null>}
    *
    * @public
    */
-  public async execute(): Promise<Dao> {
+  public async execute(): Promise<Dao|null> {
```
nivida, line 37: I think this could be a one-liner :)

packages/govern/internal/clients/graphql/queries/dao.ts
```diff
@@ -7,14 +7,20 @@ export interface Dao {
...
+
 const dao: string = `
     query DAO($name: String!) {
-      dao(name: $name) {
-        id
-        address
-        metadata
-        registryEntries {
-            ...RegistryEntry_registryEntry
+      executors: registryEntries(where: { name: $name }, first: 1) {
```
nivida: I think this is wrong.. an entry on the registry exists of Queue/Executor pair without multiple executors.

yuetloo:
Closing this PR as this has been replaced by https://github.com/aragon/govern/pull/318

PR #321, opened 2021-04-14
Create dao in Govern (Todos: see description)

yuetloo:
TODO:  
1. need to update configDefaults.ts with mainnet factory address
2. need to merge https://github.com/aragon/govern/pull/318 as the DAO queries currently fail on this branch which is based on 'develop'
3. the coverage tests for createDao is low because the tests for createDao were in mocha not jest (hardhat default).  I tried using hardhat-jest plugin, couldn't figure out how to customize it to use existing jest.config.  The plugin seems to just use its own config...

This PR addresses https://linear.app/aragon/issue/ARA2-121/create-dao-on-governjs
Note: Vocdoni part of the requirement not added here - still missing requirement on this.


yuetloo:
close this PR because it's already merged into develop by #331

PR #323, opened 2021-04-14
Feature/proposaldetails

reddyismav:
Functional HomePage, DaoPage, New proposal Flow and View Proposal.

Things not in this commit - 
- Load more not working properly.
- Contract calls.
- Footer
- Validation Checks on the New proposal page.


packages/govern-console/src/containers/HomePage/HomePage.tsx
```diff
@@ -47,12 +48,15 @@ const HomePage = ({ ...props }) => {
               updateSelectedDao={updateSelectedDaoAndPushToHistory}
             />
           </Route>
-          <Route exact path="/:daoName">
+          <Route exact path="daos/:daoName">
```
bhanusanghi, line 51: Please change the route.push method as well otherwise clicking on a DAO card will break the app.

PR #347, opened 2021-04-30
executor needs to be set before the effect is being returned.

reddyismav:
Executor was being set after a return statement causing an unreachable code. 
This was the reason which might have caused the invalid ENS name. 

I have moved it up.


bhanusanghi (approved):
Looks good

PR #351, opened 2021-05-03
Feature/dao 238 issue about createdao

Rekard0:
- fixing create dao

remaining
- moving getToken to console package
- removing current validation


novaknole:
@Rekard0 

conflict needs to be fixed

PR #360, opened 2021-05-05
fix govern test failures

yuetloo:
As a temporary workaround for the coverage test failure, I took `RegisterToken.js`, `token.js`, `Gateway.js` out of the jest coverage test.

We should add them back once we consolidated the test framework to use in govern, see https://linear.app/aragon/issue/DAO-237/convert-tests-from-jest-to-mocha-in-governjs

### DIFF
--- a/packages/govern/public/proposal.ts
+++ b/packages/govern/public/proposal.ts
@@ -1,10 +1,11 @@
 import { DaoConfig, ContainerConfig } from './createDao'
-import { ContractInterface, Contract } from '@ethersproject/contracts'
-import { Web3Provider, TransactionResponse, TransactionReceipt } from '@ethersproject/providers'
-import { BigNumberish } from '@ethersproject/bignumber'
-import { BytesLike } from '@ethersproject/bytes'
-import { toUtf8Bytes } from '@ethersproject/strings'
-import { ethers } from 'ethers';
+import {
+  ContractInterface,
+  Contract,
+  providers,
+  BigNumberish,
+  utils
+} from 'ethers'
 
 
 const Payload = `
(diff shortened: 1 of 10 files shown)


packages/govern/public/proposal.ts
```diff
@@ -65,14 +66,14 @@ export type PayloadType = {
...
-  proof: BytesLike
+  allowFailuresMap: utils.BytesLike
+  proof: utils.BytesLike
 }
 
 export type ActionType = {
   to: string
   value: BigNumberish
-  data: BytesLike
+  data: utils.BytesLike
```
novaknole, line 76: @yuetloo why does `BytesLike` come from `utils` ? and `BigNumberish` doesn't come from it ? there're many places this happens.

yuetloo:
Those are ethers.js types..

On May 6, 2021, at 4:50 AM, Giorgi Lagidze ***@***.***> wrote:

﻿

***@***.**** commented on this pull request.
------------------------------

In packages/govern/public/proposal.ts
<https://github.com/aragon/govern/pull/360#discussion_r627208619>:

>  }

 export type ActionType = {
   to: string
   value: BigNumberish
-  data: BytesLike
+  data: utils.BytesLike

@yuetloo <https://github.com/yuetloo> why does BytesLike come from utils ?
and BigNumberish doesn't come from it ? there're many places this happens.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/aragon/govern/pull/360#pullrequestreview-653148162>, or
unsubscribe

novaknole:
@yuetloo I know what they are. but for bignumberish, you don't use `utils.bignumberish` . 

This is not consistent.

yuetloo:
Doesn’t seems like bignumberish is available on utils

PR #381, opened 2021-05-12
Feature/dao 264 execution scheduling modal

yuetloo:
This PR handles user cancellation and redirect to the proposal details page on successful creating the proposals from the DAO settings page and the New Proposal page.

@novaknole , I did put in a change to extend the ExecutionTime to 5 minutes as I was getting `bad delay` error without it.. This is not related to this issue, but, without it, I would occasionally get the error, making it difficult to do testing..

here's the file:
packages/govern-console/src/utils/ERC3000.tsx, line 22.

```
    executionTime: payload.executionTime || Math.round(Date.now() / 1000) + parseInt(config.executionDelay.toString()) + 300, // add 5 minutes for network latency.

```


packages/govern-console/src/components/TransactionKeeper/TransactionKeeper.tsx
```diff
@@ -1,64 +1,29 @@
...
-});
-const TransactionListWrapper = styled('ul')(({ theme }) => ({
-  paddingInlineStart: '30px',
-}));
-const TransactionListItem = styled('li')(({ theme }) => ({
-  padding: 0,
-}));
-//* End of Styled Components
+const getErrorFromException = (ex: any): string => {
+  let errorMessage = ex.error?.message || ex.message;
```
novaknole: @yuetloo what do you think if we also add `ex.reason` beside `ex.message` ?

packages/govern-console/src/containers/DAOSettings/DAOSettings.tsx
```diff
@@ -249,8 +255,17 @@ const DaoSettings: React.FC<DaoSettingFormProps> = ({ onClickBack }) => {
         onTransactionFailure: (error) => {
           enqueueSnackbar(error, { variant: 'error' });
         },
-        onTransactionSuccess: () => {},
-        onCompleteAllTransactions: () => {},
+        onTransactionSuccess: (_, receipt: ContractReceipt) => {
+          containerHash = Proposal.getContainerHashFromReceipt(
```
novaknole, line 258: @yuetloo 

could you use `proposalInstance` instead of `Proposal`. I made `proposalInstance` as a wrapper for `Proposal` for now.. `proposalInstance` is already created in the same file.

packages/govern-console/src/containers/HomePage/ModalsContext.tsx
```diff
@@ -110,3 +110,10 @@ export const ModalsProvider: React.FC = ({ children }) => {
     </ModalsContext.Provider>
   );
 };
+
+export const closeTransactionsModalAction: CloseAction = {
+  type: ActionTypes.CLOSE,
+  payload: {
+    modal: 'transactionsModal',
+  },
+}
```
novaknole: @yuetloo  EOF error.

packages/govern-console/src/containers/NewProposal/NewProposal.tsx
```diff
@@ -450,8 +453,17 @@ const NewProposal: React.FC<NewProposalProps> = ({ onClickBack, ...props }) => {
         onTransactionFailure: (error) => {
           enqueueSnackbar(error, { variant: 'error' });
         },
-        onTransactionSuccess: () => {},
-        onCompleteAllTransactions: () => {},
+        onTransactionSuccess: (_, receipt: ContractReceipt) => {
+          containerHash = Proposal.getContainerHashFromReceipt(
```
novaknole, line 457: @yuetloo  `proposalInstance` instead of `Proposal`.

packages/govern-console/src/utils/urls.tsx
```diff
@@ -8,3 +8,16 @@
...
+ * Returns the url of the Proposal details page
+ * @param daoName the name of the DAO
+ * @param containerHash the proposal container hash
+ * @returns {string} the url of the proposal details page
+ */
+export function proposalDetailsUrl(
+  daoName: string,
+  containerHash: string,
+): string {
+  return `/proposals/${daoName}/${containerHash}`;
```
novaknole, line 22: In a perfect world, the url `/proposals/${daoName}/${containerHash}` could be changed to `${daoName}/proposals//${containerHash}`, but this will need the route change too. You can decide if you wanna do it .

packages/govern-console/src/containers/HomePage/ModalsContext.tsx
yuetloo: done

packages/govern-console/src/containers/NewProposal/NewProposal.tsx
yuetloo: @novaknole , I purposely made getContainerHashFromReceipt() a  static function so that there's no need to instantiate proposal with a provider to use this function as I think it may be useful in the situation where there's no connection to the node but we still want to get the event args.  Everything already parsed by ethers.js, just need to retrieve the value, no need for a signer/provider to get the event args here..

packages/govern-console/src/utils/urls.tsx
yuetloo: I will create a task to track the work for this.. trying to keep the changes in this PR relevant to this task only.

packages/govern-console/src/components/TransactionKeeper/TransactionKeeper.tsx
yuetloo: sure.  I'll add that.

packages/govern-console/src/containers/NewProposal/NewProposal.tsx
novaknole: Idea is that in the Govern-console, you would need this `getContainerHashFromReceipt` after the transaction has happened. otherwise, what's the point or where do you get receipt from ? So, because of this, provider will always be needed.  `onTransactionSuccess` gets called after executing a transaction so provider anyways is there.. I just said this change, because in the end, I don't want to end up having `Proposal` and `ProposalInstance` instance.  I'd follow your way but we anyways need `provider` here..

yuetloo: I agree with you that in govern-console, we only need `getContainerHashFromReceipt` after a successful transaction.  But, getContainerHashFromReceipt() is a function in Govern.js which is designed to be used by anyone. Also, we always need to have Proposal in order to make proposalInstance, I'm not duplicating anything here. It's just a different way to call the function.

PR #383, opened 2021-05-13
missing dependency fix

Rekard0:
- missing dependency fix
- adding telegram link to footer
- adding error messages to wallet, create dao and dao settings

### DIFF
--- a/packages/govern-console/src/containers/CreateDAO/CreateDaoForm.tsx
+++ b/packages/govern-console/src/containers/CreateDAO/CreateDaoForm.tsx
@@ -128,8 +128,7 @@ const CreateDaoForm: React.FC<FormProps> = ({
   const { control, handleSubmit, watch, setValue } = useForm<FormInputs>();
 
   const connectedChainId = useMemo(() => {
-    if (chainId === ChainId.RINKEBY && isConnected)
-      return ChainId.RINKEBY;
+    if (chainId === ChainId.RINKEBY && isConnected) return ChainId.RINKEBY;
 
     return ChainId.MAINNET;
   }, [chainId, isConnected]);
@@ -137,7 +136,6 @@ const CreateDaoForm: React.FC<FormProps> = ({
   // use appropriate default config
   useEffect(() => {
     const _config: DaoConfig = DEFAULT_DAO_CONFIG[connectedChainId];
-    console.log('defultconfig:', _config)
     setValue('daoConfig', _config);
   }, [connectedChainId]);
 
--- a/packages/govern-console/src/containers/CreateDAO/CreateDaoForm.tsx
+++ b/packages/govern-console/src/containers/CreateDAO/CreateDaoForm.tsx
@@ -28,7 +28,6 @@ import { validateToken } from '../../utils/validations';
 import { CreateDaoStatus } from './CreateDao';
 import { CreateDaoProgressProps } from './CreateDaoProgress';
 import { ContractReceipt } from 'ethers';
-import { useSnackbar } from 'notistack';
 
 interface FormInputs {
   /**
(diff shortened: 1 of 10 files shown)


packages/govern-console/src/containers/CreateDAO/CreateDaoForm.tsx
```diff
@@ -122,25 +123,33 @@ const CreateDaoForm: React.FC<FormProps> = ({
...
 
   // use appropriate default config
   useEffect(() => {
     const _config: DaoConfig = DEFAULT_DAO_CONFIG[connectedChainId];
+    console.log('defultconfig:', _config)
     setValue('daoConfig', _config);
   }, [connectedChainId]);
 
   const submitCreateDao = async (params: FormInputs) => {
+    if (!isConnected) {
```
novaknole: do we need this check ? having a button disabled if it's not connected is much better and enough.

packages/govern-console/src/containers/DAOSettings/DAOSettings.tsx
```diff
@@ -206,7 +205,13 @@ const DaoSettings: React.FC<DaoSettingFormProps> = ({ onClickBack }) => {
   }, [daoDetails, provider]);
 
   const callSaveSetting = async (formData: FormInputs) => {
-    console.log('formData', formData);
+    if (!isConnected) {
```
novaknole: same thing here. having a button disabled should be enough when it's not connected. no need to do this check IMO.

packages/govern-console/src/containers/DAOSettings/DAOSettings.tsx
```diff
@@ -144,7 +143,7 @@ const DaoSettings: React.FC<DaoSettingFormProps> = ({ onClickBack }) => {
   }, [daoList]);
 
   const proposalInstance = React.useMemo(() => {
-    if (provider && account && daoDetails) {
+    if (provider && daoDetails && account.address && account.signer) {
```
novaknole: any specific reason why writing `account.address && account.signer` was needed instead of `account` directly ?

packages/govern-console/src/containers/NewProposal/NewProposal.tsx
```diff
@@ -279,7 +278,7 @@ const NewProposal: React.FC<NewProposalProps> = ({ onClickBack, ...props }) => {
   const { account, provider, isConnected } = context;
 
   const proposalInstance = React.useMemo(() => {
-    if (provider && account && daoDetails) {
+    if (provider && daoDetails && account.address && account.signer) {
```
novaknole: any specific reason why writing account.address && account.signer was needed instead of account directly ?

packages/govern-console/src/containers/ProposalDetails/ProposalDetails.tsx
```diff
@@ -293,7 +293,13 @@ const ProposalDetails: React.FC<ProposalDetailsProps> = ({ onClickBack }) => {
   }, [networkName]);
 
   const proposalInstance = React.useMemo(() => {
-    if (provider && account && daoDetails && proposalInfo) {
+    if (
+      provider &&
+      daoDetails &&
+      proposalInfo &&
+      account.address &&
```
novaknole: any specific reason why writing account.address && account.signer was needed instead of account directly ?

packages/govern-console/src/containers/DAOSettings/DAOSettings.tsx
Rekard0: after having fallback provider, we can depend on account for checking in instance is creatable, and account is an object and it is never undefined, if .address and .signer are not checked they may be undefined inside account object hence breaking the app

novaknole: It didn't break before when I had `account` only. and if that's the case, we can return `account:undefined` from AugmentedWallet.

Rekard0: ok the check is moved to AugmentedWallet

PR #404, opened 2021-06-09
console environment

yuetloo:
Please review and see what I have so far is good enough to address the issue.  It's similar to how court/upgrade is setup.

I also think DEFAULT_DAO_CONFIG should be in the environment file as well so we can use different sets of contracts for staging environment.. what do you think?  Let me know if I missed anything..


novaknole:
@yuetloo 

Love the changes.  Splendid.

I agree about `DEFAULT_DAO_CONFIG`.  Also, let's take `ARAGON_VOICE_URL` there too.

packages/govern-console/src/environment/types.ts
```diff
@@ -0,0 +1,23 @@
+import { Bytes } from 'ethers';
+export type EnvironmentName = 'mainnet' | 'rinkeby' | 'staging';
+type ChainId = 1 | 4;
```
Rekard0, line 3: there is a chainId enum type at: https://github.com/aragon/govern/blob/feature/dao-392-console-environment/packages/govern-console/src/utils/types.tsx#L62

Do you think having this 2 might cause confusion down the road?

yuetloo: Good catch! I'll remove the enum ChainId as it's not used anywhere anymore.

Rekard0 (approved):

PR #407, opened 2021-06-18
Feature/dao 400 update dao setting page

Rekard0:
- update setting UI
- update create dao flow UI
- update footer UI
- update header UI


packages/govern-console/src/components/Header/Header.tsx
```diff
@@ -58,22 +55,12 @@ const Header = ({ ...props }) => {
...
-          <Button
-            onClick={props.toggleTheme}
-            icon={props.themeMode === 'light' ? <IconStar /> : <IconStarFilled />}
-          />
-          <Tag
-            label={`${networkName} network`}
-            color={theme.primaryContent}
-            background={theme.primary}
-          />
+          <StyledText name="body1">Network:{networkName.toUpperCase()}</StyledText>
```
yuetloo, line 58: textStyle("label1") is uppercase, I wonder if we should use this instead..

so is textStyle("label2"), textStyle("label3") at different fonts.

packages/govern-console/src/containers/CreateDao/CreateDaoProgress.tsx
```diff
@@ -22,9 +21,12 @@ import FailAction from './components/FailAction';
 import SuccessAction from './components/SuccessAction';
 import RegisterSuccessAction from './components/RegisterSuccessAction';
 
+declare let window: any;
+
 const CreateDaoProgress: React.FC<{
   setActiveStep: React.Dispatch<React.SetStateAction<CreateDaoSteps>>;
 }> = ({ setActiveStep }) => {
+  const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000';
   const walletContext: any = useWallet();
```
yuetloo, line 30: Think we should still keep ZERO_ADDRESS in the constant.tsx file as it can be reused. also, ethers.js library has constants.AddressZero that we can use as well.

yuetloo (approved):

yuetloo (approved):

PR #409, opened 2021-06-28
Feature/dao 399 new execution flow

yuetloo:
So, this is the first version of the new execution flow, please review.

Heads up:
The encoding of action input with complex objects does not work, i.e. arrays or objects.  I tested with string, number, address only.  I think I need to convert these field values from string to JSON objects before passing them to ethers.js encoding function to make these work.. will need to test that out..


packages/govern-console/src/containers/NewExecution/ActionList.tsx
```diff
@@ -0,0 +1,151 @@
...
+    );
+  }
+
+  return (
+    <div>
+      {actions.map((action, index: number) => (
+        <Box
+          css={`
+            & > h1 {
+              height: auto !important;
```
yuetloo, line 104: @Rekard0 this is a hack to overwrite the fixed height heading of a box in the aragon ui.  If you think we should move this to the aragon ui to allow variable length heading, let me know, i'll move it.

PR #423, opened 2021-07-05
Feature/dao 419 mobile menu

Rekard0:
- update header for mobile 
- convert help button to <a>


packages/govern-console/src/components/Wallet/Wallet.tsx
```diff
@@ -115,72 +80,76 @@ const Wallet = ({}) => {
     }
   }, [account]);
 
+  const getDisplayText = (displayText: string, isAddress: boolean) => {
+    if (isAddress) {
+      const formattedAddress =
+        displayText.slice(0, 5) +
+        '...' +
+        displayText.slice(displayText.length - 5, displayText.length - 1);
+      return formattedAddress;
```
yuetloo, line 89: there's a util function to shorten address here..
https://github.com/aragon/govern/blob/develop/packages/govern-console/src/utils/account.tsx

packages/govern-console/src/components/Wallet/Wallet.tsx
```diff
@@ -61,34 +26,34 @@ const Wallet = ({}) => {
...
         });
       } else {
-        enqueueSnackbar(error.message, {
+        toast(error.message, {
           variant: 'error',
         });
       }
     }
-  }, [status, error, enqueueSnackbar]);
+  }, [status, error, toast]);
```
yuetloo, line 57: will create a separate PR to replace all enqueueSnackbar with toast for consistent look.

packages/govern-console/src/components/Wallet/Wallet.tsx
```diff
@@ -61,34 +26,34 @@ const Wallet = ({}) => {
...
       } else if (error.message.includes('window.ethereum')) {
-        enqueueSnackbar('Please install a wallet.', {
+        toast('Please install a wallet.', {
           variant: 'error',
         });
       } else {
-        enqueueSnackbar(error.message, {
+        toast(error.message, {
           variant: 'error',
         });
```
yuetloo, line 53: is  `variant:...` leftover from `enqueueSnackbar`?

packages/govern-console/src/components/Wallet/Wallet.tsx
yuetloo: looks like `getDisplayText` is only used in 1 spot and the `isAddress` is always true, may be we can remove this function and just use the `getTruncatedAccountAddress()` function.

yuetloo (changes requested):

yuetloo (approved):

yuetloo:
approved this as @Rekard0 will address the comment in a different PR since the comment are not show stoppers..

PR #427, opened 2021-07-06
Bug fixing - DAO-32

Rekard0:
- change texts
- change logo
- bug fixing described in DAO-32


packages/govern-console/src/containers/DAOSettings/DAOSettings.tsx
```diff
@@ -202,13 +215,11 @@ const DaoSettings: React.FC<DaoSettingFormProps> = () => {
...
     };
 
-    console.log('payload', payload, 'newConfig', newConfig);
-
     if (proposalInstance) {
       try {
         transactionsQueue.current = await proposalInstance.schedule(payload, buildConfig(config));
       } catch (error) {
-        console.log('Failed scheduling dao changes', error);
+        // console.log('Failed scheduling dao changes', error);
```
githubdoramon: Why leave it commented and not removing?

packages/govern-console/src/utils/types.tsx
```diff
@@ -1,8 +1,8 @@
 export enum CustomTransactionStatus {
-  Successful,
-  Failed,
   Pending,
   InProgress,
+  Successful,
+  Failed,
 }
 
 export enum CiruclarProgressStatus {
```
githubdoramon: Not sure if it is worth to fix now, but there is a typo in this enum

packages/govern-console/src/components/TransactionKeeper/TransactionKeeper.tsx
```diff
@@ -35,17 +35,14 @@ const TransactionKeeper: React.FC<TransactionKeeperProps> = ({
 
   const updateTransaction = React.useCallback(
     (updatedTransaction: CustomTransaction, updatedTransactionIndex: number) => {
-      console.log('updating transaction' + updatedTransactionIndex);
-      const updatedTransactions = transactions.map((transaction, index) =>
-        index === updatedTransactionIndex ? updatedTransaction : transaction,
```
yuetloo, line 40: consider using the `updater` function of `SetState` to avoid race conditions when updating array state.
https://stackoverflow.com/questions/26253351/correct-modification-of-state-arrays-in-react-js
https://reactjs.org/docs/react-component.html#setstate

yuetloo: Also, can we rename `updateTransaction()` to `updateTransactionStatus(txIndex, newStatus)` and get rid of the the following code ?  This way, it's clear that we are only changing the status, nothing else in the transaction?

```
        let updatedTransaction = produce(transaction, (draft) => {
          draft.status = CustomTransactionStatus.InProgress;
        });
```

packages/govern-console/src/containers/CreateDao/CreateDaoProgress.tsx
```diff
@@ -42,22 +42,27 @@ const CreateDaoProgress: React.FC<{
...
+        // update create dao status
+        const newList2 = [...progressList];
+        newList2[1].status = CiruclarProgressStatus.Done;
+        setProgressList(newList2);
+        setShowAction('register');
       }
     };
     checkIfRegistered();
-  }, [daoTokenAddress, provider]);
+  }, [daoTokenAddress, provider, progressList]);
```
yuetloo, line 65: not sure why `checkIfRegistered()` is checked as useEffect() instead of whenever daoTokenAddress is changed...

packages/govern-console/src/containers/DAOSettings/DAOSettings.tsx
Rekard0: should be removed.
i wish we had some sort of logger so it can only log when in staging or in local and not on production

packages/govern-console/src/containers/DAOSettings/DAOSettings.tsx
```diff
@@ -164,14 +164,27 @@ const DaoSettings: React.FC<DaoSettingFormProps> = () => {
...
+      }
+    } else {
+      if (textRule !== ipfsMetadata?.text) {
+        // console.log('there is text change', textRule);
+        return await addToIpfs(textRule);
+      }
+    }
+    // console.log('no change', config.rules);
+    return config.rules;
+  };
```
yuetloo, line 178: I haven't looked in detail how the IPFSInput field stuff work yet, but, just a suggestion, may be we can use react-hook-form's formState.isDirty to detect if there's changes to the ipfs value?  This is just a suggestion, I don't know if it works..

packages/govern-console/src/utils/types.tsx
Rekard0: nice catch, thanks

packages/govern-console/src/containers/CreateDao/CreateDaoProgress.tsx
```diff
@@ -42,22 +42,27 @@ const CreateDaoProgress: React.FC<{
   const [showAction, setShowAction] = useState<'none' | 'fail' | 'register' | 'finish'>('none');
   const [rule, SetRule] = useState<BytesLike>('');
```
yuetloo, line 42: BTW, while you're on this file, can you also change line 33 to use ether's constants.AddressZero instead of defining ADDRESS_ZERO in this file?

https://docs.ethers.io/v5/api/utils/constants/#constants-AddressZero

Rekard0: perfect, thanks

packages/govern-console/src/components/TransactionKeeper/TransactionKeeper.tsx
Rekard0: yes make sense,
probably it would be better to address this and also there is more concern in TransactionKeeper,
i think a thorough refactoring

packages/govern-console/src/containers/CreateDao/CreateDaoProgress.tsx
Rekard0: i have that in mind for the next refactoring

yuetloo (approved):

yuetloo:
approved now and will address other issue later in another refactoring..

PR #433, opened 2021-07-12
Enable analytics on govern DAO-38

Rekard0:
- implement analytics
- track pages
- track user identification
- track one test event

### DIFF
--- a/packages/govern-console/public/index.html
+++ b/packages/govern-console/public/index.html
@@ -28,8 +28,9 @@
   </body>
   <script>
     !(function () {
+      const analyticsKey = '%REACT_APP_ANALYTICS_KEY%';
       var analytics = (window.analytics = window.analytics || []);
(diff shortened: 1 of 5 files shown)


packages/govern-console/public/index.html
```diff
@@ -26,4 +26,63 @@
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root"></div>
   </body>
+  <script>
+    !(function () {
+      const analyticsKey = '%REACT_APP_ANALYTICS_KEY%';
+      var analytics = (window.analytics = window.analytics || []);
+      if (!analytics.initialize && !analyticsKey.includes('REACT_APP_ANALYTICS_KEY'))
```
Rekard0, line 33: shouldn't we log a message in case the analytic was not available?

PR #441, opened 2021-07-16
Feature/dao 66

novaknole:


packages/govern-console/src/containers/HomePage/HomePage.tsx
```diff
@@ -13,13 +13,17 @@ import CreateDao from 'containers/CreateDao/CreateDao';
 import { trackPage } from 'services/analytics';
 import { useLocation } from 'react-router-dom';
 import { useEffect } from 'react';
+import scrollToTop from '../../utils/scrollToId';
```
novaknole: @Rekard0 could you import this without relative path ? the same kind of way as `services/analytics` is imported above.

packages/govern-console/src/containers/CreateDao/CreateDao.tsx
```diff
@@ -10,16 +10,15 @@ import { CreateDaoProvider } from './utils/CreateDaoContextProvider';
 import PageContent from 'components/PageContent/PageContent';
 import { PageName } from 'utils/HelpText';
 import ReviewCard from './components/ReviewCard';
+import scrollToTop from '../../utils/scrollToId';
```
novaknole: @Rekard0 could you import this without relative path ? the same kind of way as `utils/HelpText` is imported above.

Rekard0: done

PR #444, opened 2021-07-26
DAO-64 addressing comments 

Rekard0:
- Text Copy Update
- UI Bug Fixes

### DIFF
--- a/packages/govern-console/src/containers/CreateDao/CreateDaoBasicInfo.tsx
+++ b/packages/govern-console/src/containers/CreateDao/CreateDaoBasicInfo.tsx
@@ -46,10 +46,6 @@ const CreateDaoBasicInfo: React.FC<{
   const methods = useForm<ICreateDaoBasicInfo>({ defaultValues: { ...basicInfo } });
   const { control, watch, getValues, trigger } = methods;
 
-  // useEffect(() => {
-  //   setValue('isExistingToken', isExistingToken);
-  // }, [isExistingToken, setValue]);
-
   const moveToNextStep = async () => {
     const validate = await trigger();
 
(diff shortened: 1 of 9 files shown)


packages/govern-console/src/containers/CreateDao/CreateDaoBasicInfo.tsx
```diff
@@ -45,9 +43,12 @@ const CreateDaoBasicInfo: React.FC<{
...
 
-  useEffect(() => {
-    setValue('isExistingToken', isExistingToken);
-  }, [isExistingToken, setValue]);
+  const methods = useForm<ICreateDaoBasicInfo>({ defaultValues: { ...basicInfo } });
+  const { control, watch, getValues, trigger } = methods;
+
+  // useEffect(() => {
+  //   setValue('isExistingToken', isExistingToken);
+  // }, [isExistingToken, setValue]);
```
yuetloo: if this useEffect() is no longer needed, can you remove it?

packages/govern-console/src/containers/CreateDao/CreateDaoCollateral.tsx
```diff
@@ -67,6 +69,13 @@ const CreateDaoCollateral: React.FC<{
     formState: { errors },
   } = methods;
 
+  useEffect(() => {
+    if (isExistingToken) {
+      setValue('isScheduleNewDaoToken', 0);
+      setValue('isChallengeNewDaoToken', 0);
+    }
+  }, [isExistingToken, setValue]);
```
yuetloo: Can you explain why this useEffect is needed as it looks to me the form field's onChange() handler would update the values.

packages/govern-console/src/containers/CreateDao/CreateDaoBasicInfo.tsx
Rekard0: done

packages/govern-console/src/containers/CreateDao/CreateDaoCollateral.tsx
Rekard0: this use effect trigger once when component is mounted and if previouse basicInfo.isExistingToken is true.
this is needed, in the senarios of for example: isScheduleNewDaoToken set to true but user go back to basic info and change to using existing token, so when user reach this screen these variables are set to default again.

yuetloo (approved):

PR #459, opened 2021-08-13
Feature / DAO-171

Rekard0:
Feature/DAO-170
- updated dropDown component in Aragon UI
- add width to text input on Aragon UI
- implement positions for popOver of dropDown

Feature/DAO-171
- import dropDown component to Govern
- create TimeInterval Component
- implement TimeInterval in create DAO
- - implement TimeInterval in DAO Settings

### DIFF
--- a/packages/govern-console/src/containers/DAOSettings/DAOSettings.tsx
+++ b/packages/govern-console/src/containers/DAOSettings/DAOSettings.tsx
@@ -328,7 +328,7 @@ const DaoSettings: React.FC<DaoSettingFormProps> = () => {
             ) : (
               <TimeInterval
                 title="Execution delay"
-                subtitle="Amount of time any action in your DAO will be available to be challenged before bein executed."
+                subtitle="Amount of time any transaction in your DAO will be available to be disputed by your members before being executed."
                 placeholder={'Amount'}
                 inputName="delayInputValue"
                 dropdownName="delaySelectedIndex"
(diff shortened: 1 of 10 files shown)


packages/govern-console/src/containers/DAOSettings/DAOSettings.tsx
```diff
@@ -316,30 +319,24 @@ const DaoSettings: React.FC<DaoSettingFormProps> = () => {
...
+                state={CircularProgressStatus.InProgress}
+              />
+            ) : (
+              <TimeInterval
+                title="Execution delay"
+                subtitle="Amount of time any transaction in your DAO will be available to be disputed by your members before being executed."
+                placeholder={'Amount'}
+                inputName="delayInputValue"
+                dropdownName="delaySelectedIndex"
+                resultName="daoConfig.executionDelay"
```
yuetloo, line 335: I think we don't need to pass inputName, dropdownName, resultName to TimeInterval.. I think TimeInterval should create a unique name for those fields and not let caller determine that.. am I missing something?

packages/govern-console/src/utils/TimeInterval.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const intervals = TIME_INTERVALS.values;
+  let determined = false;
+  let index = intervals.length - 1;
+
+  while (!determined) {
+    if (seconds >= intervals[intervals.length - 1]) {
+      determined = true;
+      return getIntervalObject(seconds, intervals[intervals.length - 1], intervals.length - 1);
+    }
+
```
yuetloo, line 39: `the seconds > 1 day` scenario can be moved out of the while loop

yuetloo (approved):

yuetloo:
@Rekard0 , we can clean up the code regarding my comment later..

packages/govern-console/src/containers/DAOSettings/DAOSettings.tsx
Rekard0: we need to pass them as those names are the name of the variable that formProvider provides
we did it the same way for ipfs input
this is the case of nested forms

PR #470, opened 2021-09-07
Bug / DAO-206 

sepehr2github:
- add handleGetValues function to prevent sending initial undefined value for inputs


packages/govern-console/src/components/TimeInterval/TimeInterval.tsx
```diff
@@ -46,6 +46,12 @@ export const TimeInterval: React.FC<TimeIntervalInputs> = ({
     }
   }, [timeInSeconds, dropdownName, inputName, setValue]);
 
+  // prevent sending initial undefined value to inputs
+  const handleGetValues = (value: string) => {
+    const result = getValues(value);
+    return result ? result : 0;
```
Rekard0: recommendation: 
return getValues(value) || 0;

Rekard0 (approved):

PR #476, opened 2021-09-21
Bug / DAO-232

sepehr2github:
- formatTime function added 
- use formatTime function to change delay time from seconds to human readable format


WissenIstNacht (approved):
Thanks for the fix!

PR #477, opened 2021-09-22
Feature / DAO-286

Fabricevladimir:
- Set up nested routing for a given DAO's related pages
- Added skeleton page structure
- Added DaoSideCard component


packages/govern-console/src/containers/DAO/DaoHomePage.tsx
```diff
@@ -1,28 +1,151 @@
...
+  const { daoName } = useParams<any>();
+  const { pathname } = useLocation();
   const { path, url } = useRouteMatch();
+  const { layoutName } = useLayout();
+
+  /**
+   * State
+   */
+  const [daoExists, setDaoExists] = useState<boolean>(true);
+  const [daoDetails, setDaoDetails] = useState<any>();
```
WissenIstNacht: I think this is not needed (indeed, the `daoDetails` is unused). You're already getting that info from `useDaoQuery`. Is there a good reason for duplicating it?

packages/govern-console/src/containers/DAO/DaoHomePage.tsx
```diff
@@ -1,28 +1,151 @@
...
+  const [queueNonce, setQueueNonce] = useState<number>();
```
WissenIstNacht: same here?

packages/govern-console/src/containers/DAO/components/DaoSideCard/DaoSideCard.tsx
```diff
@@ -1,9 +1,119 @@
...
+  { path: '/settings', label: 'Settings', icon: <IconSettingsFilled size="small" /> },
+];
+
+const Container = styled.div`
+  border: 1px solid transparent;
+  position: relative;
+`;
+
+const AvatarContainer = styled.div`
+  position: absolute;
```
WissenIstNacht, line 38: FYI: you can (and should 😬) have props in styled-components! 🙂

For example, you could change this as follows: 

```ts
const AvatarContainer = styled.div<{ justify : string}>`
  position: absolute;
  width: 100%;
  display: flex;
  padding: 0px 24px;
  z-index: 1;
  justify-content: ${({ justify }) => justify}
`;
```

And then, in the tsx part, you could call:

```tsx
<AvatarContainer justify={layoutName === 'small' ? 'center' : 'unset'}>
```

// Just saw that you know this is a thing. Sorry for the explanation 😅.

packages/govern-console/src/containers/DAO/components/ActionLink/ActionLink.tsx
```diff
@@ -0,0 +1,46 @@
...
+  icon: React.ReactElement;
+};
+
+const ActionLink: React.FC<ActionLinkProps> = ({ url, label, icon }) => {
+  const { layoutName } = useLayout();
+
+  return (
+    <StyledDiv
+      to={url}
+      css={layoutName === 'small' ? 'justify-content: center; flex: 1' : undefined}
```
WissenIstNacht, line 38: Again, here you could extract the `layoutName === 'small'` check into a variable, and then pass it to `StyledDiv` as a prop. Based on that prop you can then move both the `justify-content` and `flex` modification to the styled component itself. 🙂 

(See `.../DaoSideCard.tsx` for more details)

packages/govern-console/src/containers/DAO/DaoMainPage.tsx
```diff
@@ -1,3 +1,4 @@
+// TODO: This component is deprecated. Remove.
```
WissenIstNacht: Please do. :D

If someone misses it they can always go back in git history. Also, if someone's indeed going to miss it (because something breaks without it), we'll be aware of it sooner!

WissenIstNacht (changes requested):

packages/govern-console/src/containers/DAO/DaoHomePage.tsx
Fabricevladimir: Good catch. This was setup anticipating that it would be passed down to one of the nested pages, specifically the actions page. All the data for that page is currently being loaded in the main page in a bid to perhaps avoid too many loading stages once the user gets to the dao's main page (`DaoHomePage`).

packages/govern-console/src/containers/DAO/DaoHomePage.tsx
Fabricevladimir: `queueNonce` will be used by the Actions page.

packages/govern-console/src/containers/DAO/components/DaoSideCard/DaoSideCard.tsx
Fabricevladimir: I just saw this today! 😄 I appreciate the explanation. It's a much better approach.

packages/govern-console/src/containers/DAO/DaoHomePage.tsx
Fabricevladimir: Right!

WissenIstNacht (approved):
Thanks for the changes :)

PR #481, opened 2021-10-12
Feature / DAO-335

Fabricevladimir:
- Added finance side-card based on new designs

### DIFF
--- a/packages/govern-console/src/containers/DAO/DaoFinancePage.tsx
+++ b/packages/govern-console/src/containers/DAO/DaoFinancePage.tsx
@@ -90,24 +90,33 @@ const ListTitle = styled.p`
 const DaoFinancePage: React.FC<Props> = ({ executorId, daoName, token: mainToken }) => {
   const toast = useToast();
   const { layoutName } = useLayout();
+
   const { provider, isConnected } = useWallet();
   const { data: finances, loading: isLoading } = useFinanceQuery(executorId);
 
   const [tokens, setTokens] = useState<FinanceToken>({});
   const [transactions, setTransactions] = useState<Transaction[]>([]);
-  const [isMediumPortrait, setMediumPortrait] = useState<boolean>(
-    layoutName === 'medium' && /portrait/.test(window.screen.orientation.type),
-  );
   const [isTransferModalOpen, setIsTransferModalOpen] = useState<boolean>(false);
 
+  // TODO: Future refactor - extract both isSmall and isMediumPortrait into one hook
   const layoutIsSmall = useMemo(() => layoutName === 'small', [layoutName]);
+  const checkIfMediumPortrait = useCallback(
+    () => layoutName === 'medium' && /portrait/.test(window.screen.orientation.type),
+    [layoutName],
+  );
+
+  const [isMediumPortrait, updateIsMediumPortrait] = useState<boolean>(() =>
+    checkIfMediumPortrait(),
+  );
+
+  const setMediumPortrait = useCallback(() => {
+    updateIsMediumPortrait(checkIfMediumPortrait);
+  }, [checkIfMediumPortrait]);
 
   useEffect(() => {
-    window.addEventListener('resize', () =>
-      setMediumPortrait(layoutName === 'medium' && /portrait/.test(window.screen.orientation.type)),
-    );
-    return () => window.removeEventListener('resize', () => null);
-  }, [layoutName]);
+    window.addEventListener('resize', setMediumPortrait);
+    return () => window.removeEventListener('resize', setMediumPortrait);
+  }, [setMediumPortrait]);
 
   useEffect(() => {
     if (!isLoading && finances) {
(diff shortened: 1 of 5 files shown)


packages/govern-console/src/containers/DAO/DaoFinancePage.tsx
```diff
@@ -64,17 +88,26 @@ const ListTitle = styled.p`
 `;
 
 const DaoFinancePage: React.FC<Props> = ({ executorId, daoName, token: mainToken }) => {
-  const { provider } = useWallet();
+  const toast = useToast();
+  const { layoutName } = useLayout();
+  const { provider, isConnected } = useWallet();
+  const { data: finances, loading: isLoading } = useFinanceQuery(executorId);
```
WissenIstNacht, line 95: Just FYI (no action required):

In Voice we introduced a type called `HookData` for hooks with this structure (as we had lot of them) which looks like this:

```ts
type HookData<T> = {
  data: T
  loading: boolean
  error: Error
}
```

packages/govern-console/src/containers/DAO/DaoFinancePage.tsx
```diff
@@ -64,17 +88,26 @@ const ListTitle = styled.p`
...
+  const [isTransferModalOpen, setIsTransferModalOpen] = useState<boolean>(false);
 
-  const { data: finances, loading: isLoading } = useFinanceQuery(executorId);
-  const context: any = useWallet();
-  const { isConnected } = context;
-  const { layoutName } = useLayout();
   const layoutIsSmall = useMemo(() => layoutName === 'small', [layoutName]);
-  const toast = useToast();
+
+  useEffect(() => {
```
WissenIstNacht, line 116: This could be extracted into it's own hook (e.g. `useSizes`), toghether with the medium portrait sate. You could even remove the useLayout call from this component and add it to `useSizes`. Then you could return a touple/object containing both boolean values (one for medium portrait and the other for small layout). If you use these size in several places I would recommend to make the change. If not, I'll leav it up to you to judge if it's useful.

packages/govern-console/src/containers/DAO/DaoFinancePage.tsx
```diff
@@ -64,17 +88,26 @@ const ListTitle = styled.p`
...
+    window.addEventListener('resize', () =>
+      setMediumPortrait(layoutName === 'medium' && /portrait/.test(window.screen.orientation.type)),
+    );
+    return () => window.removeEventListener('resize', () => null);
```
WissenIstNacht: This does not work. You have to add the exact same function (as in define it outside and pass the same reference) to the remover as you passed to the adder.

packages/govern-console/src/containers/DAO/components/FinanceSideCard/FinanceSideCard.tsx
```diff
@@ -104,13 +107,8 @@ const FinanceSideCard: React.FC<Props> = ({ tokens, mainToken, onNewTransfer })
...
   useEffect(() => {
     // Assets shown on medium to large screens
-    if (!layoutIsSmall) {
-      setDisplayAssets(true);
-    }
-
-    if (layoutIsSmall) {
-      setDisplayAssets(false);
-    }
+    if (!layoutIsSmall) setDisplayAssets(true);
```
WissenIstNacht, line 110: `setDisplayAssets(!layoutIsSmall)` :yum:

WissenIstNacht (changes requested):
Thanks mate! :)

Left some comments (some requiring actions, some not necessarily). 

Also, I have noticed some issues related to viewport on certain screen sizes. Will dm you the details with pictures etc.

packages/govern-console/src/containers/DAO/components/FinanceSideCard/FinanceSideCard.tsx
Fabricevladimir: I separated this into two if statements because I thought the logic wouldn't be as clear at first glance. Too verbose like this you think?

packages/govern-console/src/containers/DAO/DaoFinancePage.tsx
Fabricevladimir: Yeah, we have a few of them around here as well. I'll look into adding a similar type

packages/govern-console/src/containers/DAO/DaoFinancePage.tsx
Fabricevladimir: Very good point. As this is the only place I need both of them though, I'll refrain from immediately implementing a hook and maybe leave a TODO for future refactoring.

packages/govern-console/src/containers/DAO/components/FinanceSideCard/FinanceSideCard.tsx
WissenIstNacht: haha, no I'm actually very okay with it. Just found it funny you didn't go all the way :)

WissenIstNacht (approved):

PR #482, opened 2021-10-13
DAO-365: remove json.fromBytes() which can cause subgraph to fail to sync

yuetloo:
This fix will require govern console to parse the title from hex.
The problem is `json.fromBytes()` throws and causes subgraph to fail to sync. I tried `json.try_fromBytes()` and still the same situation.

I also tried upgrading the version to `graph-ts` but there are many breaking changes..


packages/govern-subgraph/src/GovernQueue.ts
```diff
@@ -66,20 +66,20 @@ export function handleScheduled(event: ScheduledEvent): void {
...
-    if(result) {
-      let data = json.fromBytes(result as Bytes)
-      payload.title = data.toObject().get('title').toString()
-    }
+  }
+
+  if (result) {
+    // log.debug('title {}', [result.toHex()])
+    // let data = json.fromBytes(result as Bytes)
+    // payload.title = data.toObject().get('title').toString()
```
Rekard0, line 81: do we need to keep the old code commented?

Rekard0 (approved):

PR #504, opened 2021-11-08
Bug/ DAO-445 

sepehr2github:
- Add deposit condition for balance validator
- add getTransactionDirection function

### DIFF
--- a/packages/govern-console/src/containers/DAO/components/ModalContents/NewTransfer.tsx
+++ b/packages/govern-console/src/containers/DAO/components/ModalContents/NewTransfer.tsx
@@ -390,7 +390,7 @@ const Transfer: React.FC = () => {
             control={control}
             rules={{
               required: 'Token amount is required.',
-              validate: validateAmount,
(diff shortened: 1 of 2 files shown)


Fabricevladimir (approved):

packages/govern-console/src/containers/DAO/components/ModalContents/NewTransfer.tsx
```diff
@@ -239,13 +239,13 @@ const Transfer: React.FC = () => {
           return result;
         }
 
-        return await validateBalance(asset, accountAddress, provider);
+        if (isDeposit) return await validateBalance(asset, accountAddress, provider);
```
githubdoramon: If it is not deposi it does not return anything? Wouldn't this block something?

sepehr2github: it shouldn't cause any issue because validator props will cover all possible values but I changed it because now looks cleaner

